PR #112, opened 2026-03-02
Add INPSource DataSource for EPANET INP files

onlyname0:
## Summary
- Adds `INPSource`, a new `DataSource` for reading EPANET INP files via WNTR
- Supports all six entity types: junctions, tanks, reservoirs, pipes, pumps, valves
- Includes geometry extraction (points for nodes, linestrings for links) and model caching
- Integrates with `DatasetCreator` via the `"inp"` source type
- Adds `wntr` as an explicit dependency

## Test plan
- [x] Unit tests for all entity types, attributes, geometry, bounding box, caching
- [x] Integration tests with `DatasetCreator` including ID linking
- [ ] Verify with a real-world INP file

### DIFF
--- a/docs/source/tutorials/data_preparation.rst
+++ b/docs/source/tutorials/data_preparation.rst
@@ -640,6 +640,88 @@ Below is given an example of a dataset creator config snippet
   }
 
 
+.. _dataset-creator-recipes-read-inp:
+
+Read an EPANET INP file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+EPANET INP files are supported via ``WNTR``. An ``inp`` source is a multi-entity source
+that exposes the six WNTR collections as separate sub-sources, each selected from the
+parent source using dot notation ``"<source_name>.<entity_type>"``:
+
+* Node types: ``junctions``, ``tanks``, ``reservoirs``
+* Link types: ``pipes``, ``pumps``, ``valves``
+
+Properties available on each sub-source match the attributes of the corresponding WNTR
+object (e.g. ``elevation``, ``diameter``, ``start_node_name``, ``base_head``). The
+synthetic ``name`` property yields the WNTR element name and is useful as a ``reference``
+attribute and as an ``id_link`` key for resolving link-to-node topology.
+
+Caveats:
+
+* The CRS can not be converted for ``inp`` data sources. The CRS must be specified to be
+  the same CRS as the coordinates stored in the INP file.
+* Link sub-sources (``pipes``, ``pumps``, ``valves``) produce line geometries from the
+  coordinates of their start and end nodes. They do not contribute a bounding box.
+* WNTR converts INP values to SI units on read according to the file's flow-units header
+  (e.g. in ``LPS``, pipe diameters expressed in millimeters become meters). Map properties
+  from the WNTR representation, not the raw INP values.
+
+Below is an example of a dataset creator config snippet. A single ``inp`` source is
+declared once, and each entity group references the appropriate sub-source through dot
+notation:
+
+.. code-block::
+
+  {
+    "__meta__": {
+      "crs": "EPSG:28992"
+    },
+    "__sources__": {
+      "network": {
+        "source_type": "inp",
+        "path": "/path/to/network.inp"
+      }
+    },
+    "data": {
+      "water_junction_entities": {
+        "__meta__": {"source": "network.junctions", "geometry": "points"},
+        "reference": {"property": "name"},
+        "water.elevation": {"property": "elevation"}
+      },
+      "water_reservoir_entities": {
+        "__meta__": {"source": "network.reservoirs", "geometry": "points"},
+        "reference": {"property": "name"}
+      },
+      "water_tank_entities": {
+        "__meta__": {"source": "network.tanks", "geometry": "points"},
+        "reference": {"property": "name"}
+      },
+      "water_pipe_entities": {
+        "__meta__": {"source": "network.pipes", "geometry": "lines"},
+        "reference": {"property": "name"},
+        "water.diameter": {"property": "diameter"},
+        "topology.from_node_id": {
+          "property": "start_node_name",
+          "id_link": [
+            {"entity_group": "water_junction_entities", "property": "name"},
+            {"entity_group": "water_reservoir_entities", "property": "name"},
+            {"entity_group": "water_tank_entities", "property": "name"}
+          ]
+        },
+        "topology.to_node_id": {
+          "property": "end_node_name",
+          "id_link": [
+            {"entity_group": "water_junction_entities", "property": "name"},
+            {"entity_group": "water_reservoir_entities", "property": "name"},
+            {"entity_group": "water_tank_entities", "property": "name"}
+          ]
+        }
+      }
+    }
+  }
+
+
 .. _tutorial-dataset-creator-config-schema:
 
 Dataset Creator Config Schema Reference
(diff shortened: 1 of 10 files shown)


elfjes (changes requested):
nice idea for using entry points! I like it :)

We do however, need to think a bit more on what the api looks like. We already have an api for plugins and it would be really good if this new entry point is consistent with that.  We have options.

we could opt to reuse the `movici.plugins` entry point for this as well. We'd create a new method for the `movici_simulation_core.core.types.Extensible` interface, such as `register_dataset_creator_source_type()`, and then let either `DatasetCreator` or `SourcesSetup` inherit from Extensible and implement that method. `DataSource` types inherit from `Plugin` and invoke that method in their implementation of `install()`. The benefit of this approach is that we make use of the existing infrastructure (such as `core.utils.configure_global_plugins`). It also means that we don't have to introduce a new name for the entry point

However, there are also drawbacks. We force every `Extensible` to try and import every `Plugin`, which may hinder start up times, such as importing aequilibrae when we want to do dataset creation (I think we do that anyway now, but it will probably make it even more difficult to change that. Although, I do see options)

---

Another approach could be to use a different name for the entry point, like we do now. I would opt for a name that is a bit more descriptive. `movici.source_type` doesn't really tell the story imho, a better name would be `movici.dataset_creator_source` or `movici.dataset_creator_source_type` or even `movici.dataset_creator.source` (<- the last one opening ourselves up for different kinds of plugins that extend `DatasetCreator`, such as certain validation steps or other shenanigans). 

Or `movici.plugins.dataset_creator` as generic dataset_creator plugin that has a few methods. We could also deprecate `movici.plugins` in favor of `movici.plugins.simulation` for equivalency. Although there might be a discrepancy with `AttributeSchema` as an Extensible that is not a `Simulation`

hmm, choices...

---

In any case, I'm not a fan of the global `_SOURCE_TYPES` dictionary. I think that it should be tied to a class, either `DatasetCreator` or `SourceSetup`. I can imagine it could be useful to have a `use_global_plugins` flag for DatasetCreator constructor and the `create_dataset` function. We don't need to register the "file' source_type and "netcdf" source type explicitly, we can do that through the plugin system

---
And perhaps this change deserves its own PR since it involves a significant architectural change

docs/source/tutorials/data_preparation.rst
```diff
@@ -640,6 +640,88 @@ Below is given an example of a dataset creator config snippet
   }
 
 
+.. _dataset-creator-recipes-read-inp:
+
+Read an EPANET INP file
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+EPANET INP files are supported via ``WNTR``. An ``inp`` source is a multi-entity source
+that exposes the six WNTR collections as separate sub-sources, each selected from the
```
elfjes: I think the source type name has changed to `"epanet"` right? That's also the name of the entrypoint

packages/drinking-water/src/movici_drinking_water_model/epanet_source.py
```diff
@@ -0,0 +1,169 @@
...
+
+    NODE_TYPES = frozenset({"junctions", "tanks", "reservoirs"})
+    LINK_TYPES = frozenset({"pipes", "pumps", "valves"})
+
+    def __init__(
+        self,
+        get_model: t.Callable[[], "wntr.network.WaterNetworkModel"],
+        entity_type: str,
+    ) -> None:
+        self._get_model = get_model
```
elfjes: might be simpler to just provide the model to the source, instead of a factory function. I don't think there's a need to delay the creation of the model, since it will definitely be created

packages/drinking-water/src/movici_drinking_water_model/epanet_source.py
```diff
@@ -0,0 +1,169 @@
...
+        self.entity_type = entity_type
+        self._items: t.Optional[t.List[t.Tuple[str, t.Any]]] = None
```
elfjes: items is a bit of an ambiguous name. I think features would be more appropriate. In general these could be named features or entities. Data sources generally have features (which is the name generally used in geospatial data). After they have been transformed into Movici data format, we call them entities. In this case, I would be fine with either.

packages/drinking-water/src/movici_drinking_water_model/epanet_source.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def _get_items(self) -> t.List[t.Tuple[str, t.Any]]:
+        if self._items is None:
+            wn = self._get_model()
+            self._items = list(getattr(wn, self.entity_type)())
+        return self._items
```
elfjes: I think it is fine to not cache the features coming from wntr and just call the wntr model function every time here. I don't think wntr does any special processing when providing an iterator over the features so it doesn't provide a performance benefit. On the contrary, duplicating them in a list adds a (slight) performance penalty

`_get_items` or `_get_features` could still be valid function that returns the iterator from the model. Or it could be a `@property`

packages/simulation-core/src/movici_simulation_core/json_schemas/dataset_creator.json
```diff
@@ -61,11 +61,7 @@
...
             "properties": {
               "source_type": {
-                "type": "string",
-                "enum": [
-                  "file",
-                  "netcdf"
-                ]
+                "type": "string"
               },
               "path": {
```
elfjes, line 66: Seems about right to not validate the source type here. in `DatasetCreator.create` we may want to validate if the name is a valid entry point, but on the other hand, we catch that quite early in `SourcesSetup` anyway

packages/simulation-core/src/movici_simulation_core/preprocessing/data_sources.py
```diff
@@ -24,6 +24,84 @@
 GeometryType = t.Literal["points", "lines", "polygons", "cells"]
 
 
+class MultiEntitySource:
```
elfjes: perhaps rename to `MutlitpleEntityTypeSource`, since every source is about multiple entities, but not all do different entity types

packages/drinking-water/src/movici_drinking_water_model/epanet_source.py
```diff
@@ -0,0 +1,169 @@
...
+        source = cls(file=source_info["path"])
+        if "entity_type" in source_info:
+            return source[source_info["entity_type"]]
+        return source
+
+    def _get_model(self) -> "wntr.network.WaterNetworkModel":
+        key = str(self.file.resolve())
+        if key not in self._model_cache:
+            self._model_cache[key] = wntr.network.WaterNetworkModel(str(self.file))
+        return self._model_cache[key]
```
elfjes: I think we don't need this model cache anymore since we only have a single `EPANetSource` instance per wntr model per dataset creator, and we can just give the model to the `_EPANetEntitySource` instances.

Question is, when do we create the wntr model? I'm not a fan of expensive constructor functions (so we shouldn't create the model in `__init__`), but I also don't think that we should delay until the `_EPANetEntitySource` instances need it for the first time. We should just give them the model. Perhaps we can do a `self.model = None` in `__init__` and fill that attribute the first time we do a `__getitem__`

elfjes (changes requested):
Nice! I like the approach with the `_EPANetEntitySource` class and only specify the source once in the dataset creator config. Much better than before :)

Some comments below

packages/simulation-core/src/movici_simulation_core/preprocessing/data_sources.py
```diff
@@ -24,6 +24,84 @@
...
+    """
+    if "." in name:
+        source_name, entity_type = name.split(".", 1)
+        try:
+            source = sources[source_name]
+        except KeyError:
+            raise ValueError(f"Source '{source_name}' not available") from None
+        if not isinstance(source, MultipleEntityTypeSource):
+            raise TypeError(
+                f"Source '{source_name}' is not a multi-entity source, "
```
elfjes:
```suggestion
                f"Source '{source_name}' is not a MultipleEntityTypeSource, "
```

elfjes (approved):
Approved with comment

PR #114, opened 2026-03-06
Add power-grid-model integration for electrical network analysis

onlyname0:
## Summary

Adds a new `power_grid_calculation` model that integrates the [power-grid-model](https://github.com/PowerGridModel/power-grid-model) library for electrical network analysis within Movici simulations.

### Supported calculations
- **Power flow** (Newton-Raphson, iterative current, linear, linear current)
- **State estimation** (with voltage, power, and current sensors)
- **Short circuit analysis** (with fault definitions)

### Architecture
- **Processor-based design**: each PGM component type (node, line, transformer, load, etc.) has a dedicated processor that handles input array construction, sparse update arrays for dynamic changes, and result writing
- **`ComponentIdManager`**: manages globally unique IDs across all PGM component types with efficient numba-backed bidirectional lookups
- **`PowerGridWrapper`**: orchestrates processors and manages the PGM model lifecycle

### Supported components
- Nodes (regular + virtual) and zero-impedance links
- Lines, cables, and two-winding / three-winding transformers
- Loads, generators, sources, and shunts
- Voltage / power / current sensors (with dynamic measurement updates)
- Faults (for short circuit) and transformer tap regulators

### Dynamic updates
- Loads and generators support dynamic P/Q updates via SUB attributes
- All three sensor types support dynamic measurement updates
- Two-phase update semantics: calculate with current state, then apply changes for next timestep
- Configurable tap changing strategy via model config

## Test plan
- [x] 40 model-level tests covering all calculation types, component types, dynamic updates, edge cases (empty groups, disabled loads), and analytical reference values
- [x] 6 unit tests for `ComponentIdManager`
- [x] All 46 tests pass after rebase onto current main
- [x] `make lint` passes (ruff, formatting, taplo)


elfjes (commented):
Not a thorough review yet, but a first impression.

Overall I think the concept looks alright, very similar approach as the WNTR integration, which is good. Aside from the comments below, I have some more remarks:

* The proposed datamodel is incompatible with the datamodel from the mv_network datasets that we already have. Now I'm not necessarily against having to port those datasets over to a new data model, but it should be a conscious choice. What do you think are be the benefits of redoing the data model like this?
* We should think about the introduction of fault entities. The thing about entities is that they have to exist in the init_data in order to be fully supported throughout the entire movici stack. Certain parts of the code can deal with new entities being introduced later but not every part (yet). From an initial glance it would seem that the existence of a fault immediately triggers the fault calculation, there is no way to "turn on (or off)" a fault. Perhaps its simpler to just add the fault properties to the connected device once the fault triggers. This can be done in an update, not necessarily in the init data. Or, what we do currently that we abstract the fault away. That is, we have a `electrical.grounded` property, that when set to a node/transformer, triggers a fault on that node. This means that we provide a somewhat nicer interface to trigger faults, but it does come at the cost of not supporting a precise specification of a certain fault. However, the question is: do we need that kind of precision?

onlyname0:
Superseded by #125.

The monorepo restructuring on main (plus the drinking-water and transport-assignment extractions) diverged too far from this branch to rebase cleanly. Since this branch made no changes to core code, #125 is a fresh extraction of the same work into a new `packages/power-grid/` workspace package, following the established pattern.

Closing this PR in favor of #125.

PR #116, opened 2026-03-17
Add support for running a simulation synchronously in a single process

elfjes:
Describe your changes
--------------------
Adds an option `Simulation(distributed=False)` to run that simulation synchronously. Instead of spawning a process for every model and service, every model runs in the same process and the models are called in sequence.

when running non-distributed, there are no update_data and init_data services, but there are comparable classes that take up this role. This is handled transparently for the models.

The flow is as following:
1. When starting up, the `Simulation` reads the distributed flag and determines whether to spawn a `DistributedSimulationRunner` (default) or a `SynchronousSimulationRunner`.
2. When spawning a `SynchronousSimulationRunner`, no Services are instantiated. All models are instantiated and wrapped in a `ModelConnector` that is configured with customized dependencies, init_data is processed by the `DirectoryInitDataClient` which was previously used only during testing, and the UpdateDataClient gets an internal store for storing updates, rather than reaching out to the update_data service. (De)serialization is disabled by introducing a `NoopSerializer`
3. The orchestrator is wrapped with a class that catches the command messages that the Orchestrator sends to the models, and these messages are consumed by the models one by one, in the order that they are defined in the scenario config. If a model's turn comes up but there is command message pending, then the model is skipped. 
4. The simulation is ready when the Orchestrator has finished, either by raising `FSMDone` in case of success, or `FSMError` in case of a failure of one or more of the models

## TODO
- [x] documentation

## Next steps
* Improve perfomance. Currently the Orchestrator waits with sending an update to a model until all its direct dependencies have finished calculating. However, when running non-distributed, only one model is busy at a time. So when the first dependency is ready, the update is queued and updates from other dependencies are not yet processed, even if they come in before the dependent model is called. In a future PR we can address this issue and only retrieve which updates to send to the dependent model, once its turn has come to calculate.
* Rename everything `InitDataHandler` and `InitDataClient` to `InitDataGetter` which I think is a more descriptive name
* I'm not quite happy with how we're dealing with the serialization strategies, but that's something for another time

Copyright license
-----------------


packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
```diff
@@ -0,0 +1,253 @@
...
+        # Would like to cast this as Orchestrator, but can't because of circular import
+        orchestrator_cls = t.cast(t.Any, service_info.cls)
+        orchestrator = orchestrator_cls()
+
+        stream = SynchronousOrchestratorStream()
+        logger = self._get_logger("orchestrator")
+
+        orchestrator.setup(settings=self.settings, stream=stream, logger=logger)
+        orchestrator.run()
+        return stream, orchestrator
```
onlyname0: `_start_orchestrator` calls `orchestrator.run()`, which internally calls `self.stream.run()`. Because `SynchronousOrchestratorStream.run()` is a noop, `orchestrator.run()` returns immediately after `fsm.start()`. This works, but the flow is non-obvious — a comment explaining why `run()` returns immediately would help future readers.

packages/simulation-core/src/movici_simulation_core/simulation/in_process.py
```diff
@@ -0,0 +1,253 @@
...
+                model = models[name]
+                orchestrator.restart_model_timer(name)
+                response = self.handle_model_command(model, command)
+                self._get_logger(name).debug(f"Sending: {response}")
+                try:
+                    stream.handle_message((name, response))
+                except FSMDone:
+                    return 0
+                except FSMError:
+                    return 1
```
onlyname0, line 222: The `while True` loop only exits via `FSMDone`/`FSMError`. If the orchestrator reaches a state where no commands are pending for any model, this becomes an infinite busy loop with no timeout or deadlock detection. Consider breaking out (or raising) if a full iteration processes zero commands.

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
```diff
@@ -0,0 +1,253 @@
...
+        update_client: SynchronousUpdateDataClient,
+        init_data_client: DirectoryInitDataClient,
+        serializer: NoopSerializer,
+    ):
+        if isinstance(model_info, ModelFromInstanceInfo):
+            model = model_info.instance
+        elif isinstance(model_info, ModelFromTypeInfo):
+            model = model_info.cls(model_info.config or {})
+        else:
+            raise ValueError("Unsupported ModelInfo")
```
onlyname0: and `distributed.py:264-270` - The `ModelFromInstanceInfo` -> use instance / `ModelFromTypeInfo` -> instantiate pattern is duplicated. Could be a shared method on `SimulationRunner` or on `ModelInfo` itself.

packages/simulation-core/src/movici_simulation_core/simulation/in_process.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+
+    def _configure_strategies(self):
+        if self.strategies is not None:
+            for strat in self.strategies:
+                strategies.set(strat)
+
+        # Ensure we have the strategies properly instantiated
+        strategies.get_instance(ExternalSerializationStrategy, schema=self.schema)
+        strategies.get_instance(InternalSerializationStrategy)
```
onlyname0, line 134: and `distributed.py:107-112` - Nearly identical strategy configuration. Could live on `SimulationRunner` base class.

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
```diff
@@ -0,0 +1,253 @@
...
+    def close(self):
+        pass
+
+    def reset_counter(self):
+        self.counter = map(str, itertools.count())
+
+
+class SynchronousSimulationRunner(SimulationRunner):
+    """A SimulationRunner that connects all models and services in a single process, as opposed to
+    the DistributedSimulationRunner that sets up models and services that run in a separate proces
```
onlyname0: typo: "separate process"

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
```diff
@@ -0,0 +1,253 @@
...
+            for name in self.model_names:
+                command = stream.pending_commands.pop(name, None)
+                if not command:
+                    continue
+                self._get_logger("orchestrator").debug(f"Sending: {command}")
+
+                model = models[name]
+                orchestrator.restart_model_timer(name)
+                response = self.handle_model_command(model, command)
+                self._get_logger(name).debug(f"Sending: {response}")
```
onlyname0: `self._get_logger(...)` is called on every iteration of the inner loop. Consider caching loggers at init time for the orchestrator and each model name.

packages/simulation-core/tests/simulation/test_simulation.py
```diff
@@ -389,3 +405,15 @@ def test_full_simulation_run(temp_output_file, tmp_path):
...
+@pytest.mark.parametrize("distributed", [False, True])
+def test_simulation_failure(tmp_path, distributed):
+    sim = Simulation(data_dir=tmp_path, debug=True, distributed=distributed)
+    sim.add_model("pub", CrashingModel({"mode": "pub"}))
+    sim.add_model("sub", SimpleModel({"mode": "pub"}))
+
+    sim.set_timeline_info(TimelineInfo(reference=0, time_scale=1, start_time=0))
+    sim.run()
+
+    assert sim.exit_code == 1
```
onlyname0, line 419: `test_synchronous.py` only tests `SynchronousUpdateDataClient`. The runner itself is only covered by the parametrized integration tests in `test_simulation.py:388-419`. Unit tests for edge cases (unsupported services error, model crash mid-update) would be valuable.

onlyname0 (changes requested):
Reusing the Orchestrator FSM rather than reimplementing scheduling logic is a good call - guarantees both modes behave the same. The split of the old `simulation.py` into the `simulation/` package is clean and makes things easier to navigate even aside from the new feature. Also like that the integration tests are parametrized over `distributed=[True, False]` - keeps both paths honest going forward.

Naming: I find `SynchronousSimulationRunner` / `distributed=False` a bit misleading - both modes are "synchronous" from the caller's side. Something like `InProcessSimulationRunner` / `in_process=True` would better describe what's actually different. Worth getting right before merging since this is public API.

packages/simulation-core/src/movici_simulation_core/simulation/in_process.py
elfjes: This is handled in the orchestrator state machine, it ensures that when no model is pending, we either move to the next time, or in case no model has a pending next time, we go into finalization. Once we are in finalization, we let every model close itself, which triggers a model response. When the last model has closed/terminated, the orchestrator always exits with either an FSMDone or FSMError in case of an error

packages/simulation-core/src/movici_simulation_core/simulation/in_process.py
elfjes: Yes, I'm not happy with how we configure the serialization strategies. But that's for another PR

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
elfjes: I've introduced an `Orchestrator.start()` method and use that instead of its `run()` method. Better?

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
elfjes: Moving it to the Info objects themselves makes sense

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
elfjes: thnx

packages/simulation-core/tests/simulation/test_simulation.py
elfjes: Added some more unit tests. A model crashing mid-update is exactly the test case for that `test_simulation_failure` test

elfjes:
Thanks, I've processed your suggestions. I've renamed the SynchronousSimulationRunner to InProcessSimulationRunner (and its associated names). I'm happy with `distributed` as a name, and we're lucky that `distributed=True/False` is the only public facing api of this thing.

docs/source/simulations/settings.rst
```diff
@@ -1,2 +1,29 @@
+.. include:: ../include/glossary.rst
+.. include:: ../include/code.rst
+
 Settings
 ===================
+
+distributed
+-----------
+By default, a |Simulation| runs distributed. That is, every |Model| and every |Service| runs in
+its own process. This has the benefit for certain simulations, some models may calculate parallel
```
onlyname0: calculate in parallel

docs/source/simulations/settings.rst
```diff
@@ -1,2 +1,29 @@
...
+which speeds up the simulation. It is also a requirement for (future) support of running a
+Simulation on multiple nodes, where certain models may run on different sites. However, it also
+has drawbacks. It leads to communication overhead and some models may not function properly in
+a multiprocessing environment on certain operating systems (most notably the
+``traffic_assignment`` model on MacOS). To provide an alternative calculation environment, a
+|code_Simulation| can be instantiated with the ``distributed=False`` setting. Running in a
+non-distributed/in-process environment has the following benefits:
+
+- Reduce overhead. Data is not serialized/deserialized between models. Also, in a distributed the
```
onlyname0: missing a noun, should be something like "in a distributed setup, the number of processes"

packages/simulation-core/src/movici_simulation_core/services/orchestrator/service.py
```diff
@@ -70,13 +71,24 @@ def send(message: Message):
...
                 return 1
             return 0
 
+    def start(self):
+        """Start the orchestrator FSM but do not run the stream (yet)"""
+        self.fsm.start()
+
+    def restart_model_timer(self, model: str):
+        """Allow resetting a specific model timer. This is used when running using the
+        InProcessSimulationRunner. Orchestrator starts the time by default when it sends out the
```
onlyname0: starts the timer

packages/simulation-core/src/movici_simulation_core/simulation/distributed.py
```diff
@@ -0,0 +1,295 @@
...
+        * create the service
+        * create a (router) socket
+        * announce the port
+        * run the Service
+        * raise exception on failure
+
+    * Fills the ServiceInfo object
+    * Raising an exception if it fails to announce the port in time
+
+    By creating the process as deamon=True, services cannot spawn their own subprocesses but they
```
onlyname0: shouldn't it be daemon ?

packages/simulation-core/src/movici_simulation_core/simulation/distributed.py
```diff
@@ -0,0 +1,295 @@
...
+
+        * create the model with its model adapter
+        * create a (dealer) socket
+        * run the model
+        * catch exceptions from model, send ERROR message
+        * raise exceptions when not directly from model
+
+    * Fills the ModelInfo object
+
+    By creating the process as deamon=False, models can spawn their own subprocesses
```
onlyname0: again daemon ? English can be weird

packages/simulation-core/src/movici_simulation_core/simulation/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+    def configure(self, config: dict):
+        """Configure a simulation by scenario config. All model types and additional services that
+        are present in the simulation must first be registered as a plugin (see `Simulation.use`).
+        """
+        self.settings.apply_scenario_config(config)
+
+    def add_model(self, name: str, model: t.Union[Model, t.Type[Model]], config=None):
+        """
+        Manually add a model to a Simulation. A model can be added as an instance, or as
+        class. When added as a class, instantiation is of the model is done inside its subprocess,
```
onlyname0: instantiation of the model is done

packages/simulation-core/src/movici_simulation_core/simulation/__init__.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+        elif isinstance(model, Model):
+            self.active_modules[name] = ModelFromInstanceInfo(name, daemon=False, instance=model)
+        else:
+            raise TypeError(f"Invalid model type '{model.__class__}")
+        self.schema.add_attributes(model.get_schema_attributes())
+
+    def set_timeline_info(self, timeline_info: TimelineInfo):
+        """
+        When configuring the Simulation manually, use this  method to add timeline information
```
onlyname0: double space in "use this method"

docs/source/simulations/settings.rst
```diff
@@ -1,2 +1,29 @@
...
+  number of processes that are created can sometimes reach 50+ for certain scenarios. This is
+  significant, especially when you then also want to run multiple Simulations in parallel.
+- Provide determinism. When running models in parallel, there is no guarantee that a reproducible
+  result is reached. While in general a simulation will always converge to a single state, this
+  is by no means a given, and certain race conditions could occur that lead to non-deterministic
+  effects. By ensuring that only a single model runs at a time and the models are invoked in a
+  prescribed order, the simulation can be carried out deterministically, provided that the models
+  themselves are deterministic
```
onlyname0: missing . after deterministic

packages/simulation-core/src/movici_simulation_core/simulation/synchronous.py
onlyname0: yep looks good

PR #119, opened 2026-03-20
Move pre-commit from project dependencies to dev group

onlyname0:
## Summary
- Add pre-commit to `[dependency-groups] dev`

## Test plan
- [x] `uv sync` resolves successfully
- [x] Pre-commit hooks still run on commit

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = ["movici-simulation-core"]
 [dependency-groups]
 dev = [
   "movici-simulation-core[all]",
+  "pre-commit",
   "pytest",
   "pytest-cov",
   "mypy",
--- a/uv.lock
+++ b/uv.lock
@@ -225,6 +225,15 @@ wheels = [
 ]
 
 [[package]]
+name = "cfgv"
+version = "3.5.0"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


elfjes (commented):
Hmm I'm not sure what I think about this.

In general I think a tool like this should be installed separately and not as a project dependency (we're also not specifying `git` and `make` as a dependency). However, there is something to be said to add it as a dependency. The difference between this and `git` or `make`, is that this one is readily available on pypi and the other two aren't

However, there is a complication. After installation, it can (by default) only be invoked by running `uv run pre-commit`. To run it as just `pre-commit`, you'd first need to activate the venv explicitly (`source .venv/bin/activate`), uv doesn't do that for you.

So installing the git hook would be for example `uv run pre-commit install`. Which runs just fine. However, the git hook will now by default fail, because it doesn't run through uv. So you'd be forced to always activate the venv. One of the beauties of uv is that you don't explicitly have to actvivate the venv if you invoke the commands through `uv run`.

So while we can add this tool as a dependency, I don't think it provides a better UX/DX. One reason that I can think of for adding it as a dependency is in case we use new (or very old) pre-commit features, so that we'd want to pin the version, but in general any pre-commit version works; it's the hooks that need to be versioned (which we do in .pre-commit-config.yaml`). However, sharing the environment with the code environment may also lead to a dependency confict at some point. So it's not just beneficial.

What we can do, is to update the readme and just tell them to install using `uv tool install pre-commit` which will install it globally. If we have a specific version that needs be installed, we can adjust that instruction.

PR #120, opened 2026-03-24
Extract drinking water model into separate workspace package

onlyname0:
## Summary

- Move the drinking water model (model, attributes, dataset, network wrapper, JSON schema)
  from `movici-simulation-core` into a new `movici-sim-drinking-water` workspace package
  under `packages/drinking-water/`
- Register the model and attributes via `movici.plugins` entry points so plugin discovery
  continues to work without a direct import from simulation-core
- Extract shared test fixtures (`create_model_tester`, `global_schema`, `clean_strategies`,
  etc.) into a `movici_testing` pytest plugin (via `pytest11` entry point) so both packages
  can reuse them without duplication
- Add `movici-sim-drinking-water` as a root workspace dependency

## Test plan

- [ ] Run `uv run pytest packages/drinking-water/tests` — drinking water tests pass in the new package
- [ ] Run `uv run pytest packages/simulation-core/tests` — simulation-core tests still pass without the drinking water code
- [ ] Verify `uv sync --all-groups` resolves cleanly


packages/drinking-water/tests/test_model.py
```diff
@@ -813,7 +813,7 @@ def test_valve_initial_status(self, tester, valve_open):
         assert (abs(valves["drinking_water.flow"][0]) > 1e-10) is expected_open
 
     @pytest.mark.parametrize("valve_open", [False])  # a single parametrization to change fixture
-    def test_valve_status_change_reopens(self, tester):
+    def test_valve_status_change_reopens(self, tester, valve_open):
```
elfjes, line 816: curious, how did we not notice this earlier? You'd think that the test throws an error on a missing argument

packages/drinking-water/pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+drinking_water_attributes = "movici_sim_drinking_water.attributes:DrinkingWaterNetworkAttributes"
+
+[tool.hatch.version]
+source = "uv-dynamic-versioning"
+
+[tool.uv-dynamic-versioning]
+fallback-version = "0.0.0"
+
+[tool.ruff]
+extend = "../../pyproject.toml"
```
elfjes: I'm not sure if this is required. I think ruff by defaults finds the parent pyproject.toml, but I might be wrong

packages/drinking-water/pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[tool.ruff.lint.per-file-ignores]
```
elfjes: for now we've put all ruff stuff in the root pyproject.toml. I think that keeps things more together.

elfjes (commented):
Nice work! Some comments:
 
* I don't think the "sim" in the package name makes much sense. I think either "movici_drinking_water" or "movici_drinking_water_model" would be preferable. And I think a good convention would be that the package directory is the same as the package name except for the `"movici_"` prefix and using hyphens instead of underscores
* there's no reason for the tests to be nested into models/drinking_water. Putting them directly under tests is fine
* nitpick: `json_schemas` directory only has a single schema so may be renamed to just `json_schema`. Perhaps we can simplify and throw the schema just in the package root directory, we then don't need to export a `SCHEMA_PATH` variable, but can just build the schema path in the `model.py` itself
* I think we need to add a README.rst to the package, so that when we publish to pypi, we provide at least some basic info.

We also need to make sure that we run the tests in CI, and provide a better DX for running all tests during development. Perhaps we need to move the testing directives from the movici simulation core Makefile into the root Makefile, and make sure that the commands work for all packages. May need to give special attention to the coverage command.

packages/drinking-water/README.rst
```diff
@@ -0,0 +1,29 @@
...
+
+Features
+--------
+
+- Hydraulic simulation of water distribution networks
+- Support for junctions, tanks, reservoirs, pipes, pumps, and valves
+- Pressure-dependent demand (PDD) modelling
+- Pause/restart simulation pattern for integrating external control changes
+- CSR curve data for pump head curves and tank volume curves
+
```
elfjes: These three features seem like a strange choice to mention:

* PDD, good to mention, but then you should also mention DDD
* pause/restart is required for any (time dependent) Movici model
* CSR curve data is very much an implementation detail

packages/drinking-water/pyproject.toml
```diff
@@ -3,12 +3,30 @@ requires = ["hatchling", "uv-dynamic-versioning"]
...
+  "License :: Other/Proprietary License",
+  "Operating System :: Microsoft :: Windows",
+  "Operating System :: POSIX :: Linux",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Topic :: Scientific/Engineering",
+  "Topic :: Scientific/Engineering :: GIS",
```
elfjes: This model does not have much to do with GIS

Makefile
```diff
@@ -7,6 +7,19 @@ toml-check:
 
 lint: ruff toml-check
 
+unittest:
+	NUMBA_DISABLE_JIT=1 uv run pytest -v packages/*/tests/
+
+coverage:
+	NUMBA_DISABLE_JIT=1 uv run pytest \
+		$(patsubst %,--cov %,$(wildcard packages/*/src/*)) \
```
elfjes, line 15: nice oneliner

packages/drinking-water/Makefile
```diff
@@ -0,0 +1,23 @@
...
+test-numba:
+	uv run pytest -v tests/
+
+mypy:
+	- mypy $(MODULE_NAME)
+
+bump-version:
+	bumpversion  --config-file .bumpversion.app $(level)
+	@NEW_VERSION=$$(tail -1 VERSION);\
+	echo New version: $$NEW_VERSION
```
elfjes: Now that we're on uv + hatchling, we use dynamic versioning so we don't need bump-version any more

elfjes (changes requested):
Nice, I think we're almost there

packages/drinking-water/README.rst
onlyname0: Is the following better:
- Hydraulic simulation of water distribution networks using WNTR
- Demand-driven (DDA) and pressure-dependent demand (PDA) analysis
- Support for junctions, tanks, reservoirs, pipes, pumps, and valves
- Pressure, head, flow, velocity, and demand output per network element

packages/drinking-water/pyproject.toml
onlyname0: removed

packages/drinking-water/Makefile
onlyname0: removed

packages/drinking-water/README.rst
elfjes: seems good :)

elfjes (approved):
looks good
/approve

PR #121, opened 2026-03-31
Extract transport assignment model into separate workspace package

onlyname0:
## Summary
- Extract the aequilibrae-dependent traffic assignment model and corridor model into a new `packages/transport-assignment` workspace package (`movici-transport-assignment-model`)
- Remove the corridor model from the core package (it was aequilibrae-dependent)
- Move shared test fixtures into `movici_simulation_core.testing.fixtures` for reuse across packages
- Add package-level Makefile, README, pyproject.toml with entry points and classifiers

Depends on #120.

## Test plan
- [ ] `make unittest` passes from root (runs all package tests)
- [ ] `make coverage` passes from root
- [ ] Transport assignment tests pass in isolation: `cd packages/transport-assignment && make unittest`
- [ ] CI workflow runs successfully


packages/transport-assignment/tests/conftest.py
```diff
@@ -0,0 +1,346 @@
...
+            "simulation_info": {
+                "reference_time": 1_577_833_200,
+                "start_time": 0,
+                "time_scale": time_scale,
+                "duration": 730,
+            },
+            "models": [model_config],
+        },
+        "init_data": init_data,
+    }
```
elfjes: These seem like generic fixtures that we can move to the pytest plugin in simulation-core? or is there some reason that we can't do that?

packages/transport-assignment/src/movici_transport_assignment_model/json_schemas/__init__.py
```diff
@@ -0,0 +1,5 @@
+import pathlib
+
+SCHEMA_PATH = PATH = pathlib.Path(__file__).parent
```
elfjes: what's the purpose of `PATH` here?

elfjes (changes requested):
Overall looks good. Some minor requests. 

* can we rename the `integrations` directory to `ae_integration`?
* see comments below

packages/transport-assignment/tests/conftest.py
onlyname0: Good call, the generic ones (`model_name`, `time_scale`, `init_data_handler`, `add_init_data`, `config`, `global_timeline_info`) are already duplicated in `simulation-core/tests/models/conftest.py` so they should indeed move to the core pytest plugin. The aequilibrae fixtures (`clean_aequilibrae_project`, `patch_aequilibrae`) need to stay here though since core shouldn't depend on aequilibrae.

packages/transport-assignment/src/movici_transport_assignment_model/json_schemas/__init__.py
onlyname0: No purpose — it's a legacy alias copied from the core `json_schemas/__init__.py`. Only one file in the entire codebase still imports `PATH`; everything else uses `SCHEMA_PATH`. Will drop it here.

elfjes: ok :+1:

packages/transport-assignment/tests/conftest.py
elfjes: sounds good

packages/transport-assignment/tests/test_ae_project.py
```diff
@@ -7,12 +7,15 @@
...
+from movici_transport_assignment_model.integrations.collections import (
     AssignmentResultCollection,
     LinkCollection,
     NodeCollection,
 )
-from movici_simulation_core.integrations.ae.project import AssignmentParameters, ProjectWrapper
+from movici_transport_assignment_model.integrations.project import (
+    AssignmentParameters,
+    ProjectWrapper,
+)
```
elfjes, line 18: nitpick: trailing comma has changed formatting. remove?

elfjes (approved):
approved with nitpick 👍

packages/transport-assignment/tests/test_ae_project.py
onlyname0: ruff forces this style due to line length

PR #122, opened 2026-04-04
Remove deprecated support for components

elfjes:
Describe your changes
--------------------
We have deprecated and stopped using components long time ago. Let's properly remove support and clean things up

Copyright license
-----------------


onlyname0 (commented):
Nice cleanup — the tuple-form removal is long overdue. A few notes:

**Blocking**

- `dumps` return annotation is `-> bytes`, but the JSON branch returns `str` from `json.dumps`. The prior `-> str` was wrong for msgpack; the new annotation just flips which branch is wrong. Suggest `t.Union[str, bytes]`, or splitting into two methods.

**Non-blocking**

- The `# spec cannot be None ...` comment + `assert` encode an invariant that `get_spec`'s signature doesn't express. An `@t.overload` on `get_spec` that returns `AttributeSpec` (non-optional) when `default_data_type is not None` would let you drop both the comment and the assert.
- Consider a small parametrized test covering the new `TypeError("name must be a single string")` branch — the deleted `test_schema_legacy_invalid` had coverage for non-str inputs that isn't replaced.
- Minor: the previous JSON path in `dumps` called `self.dump_dict(dataset)` twice; assigning to `list_data` and reusing it is a nice redundancy cleanup (not a behavior fix).
- The final `raise` arms in `loads`/`dumps` are unreachable given `supported_file_type_or_raise` upstream, but fine as type narrowers.

PR #123, opened 2026-04-16
Add data-core package for a movici data api server

elfjes:
Describe your changes
--------------------

Traditionally, when running Movici locally, we have stored and retrieved data from (json) files on disk. While this is simple, it also has very poor performance when interacting with that data, for example when running Movici Flow against that data (using movici-viewer). We have previously expressed the wish to be able to use a database (eg sqlite) for storing data when working with Movici locally and have had some progress in that (#60). However, Movici Viewer does not support working with a sql database yet. At the same time, the Movici Cloud Platform has a data engine that serves movici data backed by a sql database. However, that code has no-longer-maintained dependencies and is furthermore not available to external users. In the light of "let's kill two birds with one stone", this PR implements a backend for api servers to configure and use a sql database for storing and serving movici data from.

The idea is to have a common base for different use cases. This common base contains:
* logic for interacting with the database
* domain logic on how to process certain tasks
* a configurable http api surface for serving out data

The following yaml describes the constraints for the data core
```yaml
general:
  goals:
    - |
      The goal for the design of this movici database model is to provide a single structure for storing data related to
      movici simulations. This structure should accomodate flexible usage for local usage of Movici, but also running a 
      multiuser database in the Movici cloud.
    - |
      Ensure forwards compatibility. Especially important for local databases (SQLite). Database files may reside on 
      a file system for many years. New versions of Movici must be able to detect a particular schema version and work
      withk that version, preferably without having to migrate the database schema.
  non_goals:
    - |
      provide authentication and authorization. These may be be added by individual applications working with this data core
      model. However, we may want to support being able to inject authorization logic in order to performantly allow or
      disallow certain operations
    - |
      data versioning will not be considered for the initial implementation. A future version of this data model may
      implement versioning to support multiple runs of the same scenario, and updates to initial datasets that do not
      affect historical simulations
  constraints:
    - |
      "Support different (SQL) database technologies: SQLite, PostgresSQL and MySQL"
    - Support movici data dir as a file backend.
    - Used by Movici Viewer, DataCollector when in sqlite mode, Data Engine (cloud platform)
design:
  - |
    The data model supports operating in different modes and with different flags to accomodate the various use cases.
    These are explained below.
  - |
    dataset_types, entity_types and attribute_specs are defined in the database and are used as foreign keys, so that
    changes to a name are reflected in the data and model configurations
  - |
    model types are stored in the database along with a config schema, so that scenarios using that model may be
    validated
  - |
    entity data is stored as raw numpy arrays, where the numpy type, shape and raw data are stored together in a single
    row
  - |
    the data section of unstructured (json) datasets are stored in a separate table. Data is split into chunks of 100MB
    that are store as blobs. An "encoding" and a "compression" field is attached, so that different encoding and or
    compression algorithms may be used (json, msgpack, gz, etc...)
  - |
    binary data is split into chunks of 100 MB that are stored as blobs. Before storing the data, the data's mimetype
    should inferred from the extension and/or file upload parameters and compared against the dataset_type's allowed
    mimetype. A "compression" field is attached, so that different compression algorithms may be used
  - |
    The domain model is kept separate from the SQL model. SQLalchemy model classes may not escape the SQLAlchemy
    repository. The domain model contains the main classes that represent domain objects. This may be dataclasses
    or pydantic models
  - |
    async first for maximum compatibilty. It is easier to run coroutines in a sync way than vice-versa (without
    sacrificing the benefits of asyncio)

modes:
  - name: SINGLE_SCENARIO
    description: |
      This mode is designed for maximum portability and flexibility. A scenario, its datasets and its results are all
      stored together in a single database that can be distributed as the user pleases.
    features:
      - a default workspace "__default__" is created that is used for every scoped resource
      - every dataset is automatically added to the scenario
      - all STRICT* flags are off by default

  - name: SINGLE_WORKSPACE
    description: |
      This mode is designed for single-user use. A user can store all their datasets and scenarios in a single database.
      This database can be tied to a specific project but doesn't have to be.
    features:
      - a default workspace "__default__" is created that is used for every scoped resource
      - scenarios and datasets must be linked explicitly
      - all STRICT* flags are off by default

  - name: MULTIPLE_WORKSPACES
    description: |
      This mode is designed for supporting a multi-user environment with multiple workspaces. Since multiple users will
      be working together, it is important that data is consistent and user input is well defined and validated
    features:
      - no default workspaces are created, these must be created before any other data can be added
      - scenarios and datasets must be tied to a workspace
      - scenarios and datasets must be linked explicitly
      - all STRICT* flags are on by default

flags:
  - name: STRICT_ATTRIBUTES
    features:
      when_set:
        - attributes and their data type must be defined in the database before they can be added to the database
      when_unset:
        - the data type of attributes may be supplied by the dataset/update when uploaded. If provided the attribute
          definition will be checked against any other attributes in the database. If no other definition for that
          attributed was found, it will be added to the database. If a conflicting definition was found, the upload is
          rejected. Otherwise the upload is accepted

  - name: STRICT_DATASET_TYPES
    features:
      when_set:
        - dataset_types must be defined before dataset data can be created
      when_unset:
        - dataset_type is automatically created for a dataset if it doesn't exist
        - |
          It tries to infer the format. 
          For json datasets, the format must be either "entity_based" or "unstructured", this must be present in the dataset
          For other datasets, the format will be set to BINARY
        - If the type already exists, the provided data will be checked against the type

  - name: STRICT_ENTITY_TYPES
    features:
      when_set:
        - entity_types must be defined before entity groups can be created
      when_unset:
        - entity_type is automatically created for an entity group if it doesn't exist

  - name: STRICT_MODEL_TYPES
    features:
      when_set:
        - models types must be defined before a scenario can be created
      when_unset:
        - a model definition will be created for every model that does not exist
        - a pass-all model validation jsonschema will be created for these models
        - this does mean that for those models, dataset-, entity-type- and attribute names are not extracted from the
          scenario model configs, so the model config may run out of sync with the data in the database

  - name: STRICT_SCENARIO_DATASETS
    features:
      when_set:
        - require that all datasets attached to a scenario exist prior to creating/updating the scenario
      when unset:
        - autogenerates a dataset stub when a scenario is created/updated with a non-existing dataset
```

Now, the implementation contains the following main layers:

* domain layer: provide the domain objects and domain logic for working with movici data that is independent from the storage method (database or file storage)
* service layer: provide some additional logic between the domain and the repository. The service layer should be independent of the repository so that it can be reused when using a file-based storage.
* repository layer: Provide the logic for converting between domain objects and database data (using SQLAlchemy)
* api layer: provide basic functionality for running an api server. Includes validation, endpoints and a way to configure/compose those endpoints into an application (FastAPI + Pydantic)

The current domain model is using dataclasses. we may want to switch to pydantic for the domain model as well, but only if the representation of the domain objects is similar enough to the representation of those objects in the api layer.

Copyright license
-----------------


packages/data-core/src/movici_data_core/database/general.py
```diff
@@ -0,0 +1,107 @@
...
+    options_count = (await session.scalar(select(func.count(Options.id)))) or 0
+
+    if metadata_count > 0 or options_count > 0:
+        raise DatabaseAlreadyInitialized
+    await session.execute(insert(Metadata).values(id=1, version=DEFAULT_SCHEMA_VERSION))
+
+    workspace_id = None
+    scenario_id = None
+    if mode in (DatabaseMode.SINGLE_SCENARIO, DatabaseMode.SINGLE_WORKSPACE):
+        workspaces_count = (await session.scalar(select(func.count(Options.id)))) or 0
```
onlyname0, line 49: **Copy-paste bug.** The variable is named `workspaces_count` but the query counts `Options.id`. Today this is a redundant no-op (the earlier guard already rejected `options_count > 0`), but the intent was clearly to check for pre-existing workspaces. Should be `select(func.count(Workspace.id))`.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -0,0 +1,152 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
+
+from . import model as db
+from .general import get_options
+from .repository import SQLAlchemyRepository
+
+
+class SQLAlchemyBackendFactory:
+    def __init__(self, session_factory: async_sessionmaker[AsyncSession]):
+        self.session_factory = session_factory
```
onlyname0: Stray `pass` after the assignment — remove.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -0,0 +1,152 @@
...
+            return SQLAlchemyBackend(
+                session,
+                options,
+                workspace_id=workspace_id,
+                scenario_id=scenario_id,
+                single_workspace_mode=True,
+                single_scenario_mode=True,
+            )
+
+        assert False, f"Unknown database mode {options.mode}"
```
onlyname0, line 128: `assert` is stripped under `python -O`. Use `raise ValueError(f"Unknown database mode {options.mode}")` so the guard survives optimized runs.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -0,0 +1,751 @@
...
+
+    def to_domain(self) -> domain_model.Dataset:
+        return domain_model.Dataset(
+            id=self.id,
+            name=self.name,
+            display_name=self.display_name,
+            dataset_type=self.dataset_type.to_domain(),
+            workspace=self.workspace.to_domain(),
+            general=self.general,
+            espg_code=self.epsg_code,
```
onlyname0: Typo `espg_code` in the domain `Dataset` dataclass (see `domain_model.py:137`). The SQL column is `epsg_code`; rename the domain field to match so this bridge can be `epsg_code=self.epsg_code`.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -0,0 +1,751 @@
...
+    dataset: Mapped[Dataset] = relationship()
+    attribute: Mapped[Attribute] = relationship()
+
+
+class RawData(Base):
+    __tablename__ = "raw_data"
+    id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=uuid.uuid4)
+    dataset_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("dataset.id", ondelete="CASCADE"))
+    encoding: Mapped[str | None]
+    compresion: Mapped[str | None]
```
onlyname0, line 294: Typo — should be `compression`. This becomes a table column name; far cheaper to fix before the schema ships than after.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -0,0 +1,751 @@
...
+
+    id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=uuid.uuid4)
+    update_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("update.id", ondelete="CASCADE"))
+    attribute_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("attribute.id", ondelete="CASCADE"))
+
+    update: Mapped[Update] = relationship()
+    attribute: Mapped[Attribute] = relationship()
+
+
+# class SimulationDatabase:
```
onlyname0: ~300 lines of commented-out legacy `SimulationDatabase` starting here. Please delete before merge — git history preserves it.

packages/data-core/src/movici_data_core/database/repository.py
```diff
@@ -0,0 +1,1185 @@
...
+                    db.ModelType.name.in_(model_types),
+                )
+            )
+        }
+        to_create = []
+        for model_type in model_types:
+            if model_type not in existing_model_types:
+                if self.options.STRICT_MODEL_TYPES:
+                    raise ResourceDoesNotExist("model_type", name=model_type)
+                to_create.append(model_type)
```
onlyname0: Duplicates in the input (a scenario with two models of the same yet-unseen type) land in `to_create` twice and the bulk insert then hits the UNIQUE constraint on `ModelType.name`. Dedupe before the insert — e.g. `to_create = list(dict.fromkeys(to_create))`.

packages/data-core/src/movici_data_core/database/repository.py
```diff
@@ -0,0 +1,1185 @@
...
+                )
+                .returning(db.Update.id)
+            ),
+        )
+        await _EntityDataHandler(self.session, self.all_data, UpdateDataSelector()).store(
+            update_id, obj.data
+        )
+        return update_id
+
+    async def delete_all(self):
```
onlyname0: **Orphan rows.** FK cascades go `Attribute → UpdateAttribute → Update` (downward), so deleting an `Update` cascades `UpdateAttribute` but leaves the `Attribute`, `DataArray` and `RowptrArray` rows created in `_EntityDataHandler.store` orphaned. `DatasetDataRepository.delete` handles this explicitly via `delete(db.Attribute).where(... in select(db.DatasetAttribute.attribute_id)...)` — mirror that here (and note that scenario deletion hits this path too via cascade). Worth a regression test that asserts `select(count(Attribute.id)) == 0` after a scenario delete.

packages/data-core/src/movici_data_core/database/repository.py
```diff
@@ -0,0 +1,1185 @@
...
+        )
+
+    async def _store_rowptr_array(self, arr: np.ndarray, attribute_id: UUID):
+        await self.session.execute(
+            insert(db.RowptrArray).values(data=arr.tobytes(), attribute_id=attribute_id)
+        )
+
+    # TODO: We already expect to have numpy data here, so we need to move this function somewhere
+    # else. The backend maybe?
+    def _data_as_numpy_dict(self, data: dict, filetype: FileType, data_is_numpy=False):
```
onlyname0: Dead and broken — `self.schema` and `self.serializer` are never set on `_EntityDataHandler` (which only assigns `session`, `all_data`, `selector`). The empty docstring and TODO above confirm it's a leftover. Remove.

onlyname0 (commented):
Thorough pass on the data-core package. Overall the layering (domain / service / repository / backend) is clean and coverage is broad — leaving inline notes on correctness issues, one orphan-row concern in the update path, and a few pre-merge cleanups.

packages/data-core/src/movici_data_core/domain_model.py
```diff
@@ -0,0 +1,155 @@
...
+    name: str
+    display_name: str
+    description: str
+
+    epsg_code: int
+    bounding_box: tuple[float, float, float, float] | None = None  # minx miny maxx maxy
+    simulation_info: dict = dataclasses.field(default_factory=dict)
+    status: ScenarioStatus = ScenarioStatus.READY
+
+    id: UUID | None = None
```
onlyname0, line 129: Other domain dataclasses (`Workspace`, `DatasetType`, `EntityType`, `AttributeType`, `ModelType`, `ScenarioDataset`) declare `id` with `dataclasses.field(compare=False, default=None)`. `Scenario` and `Dataset` don't — so equality silently depends on ids matching, and `test_scenario_round_trip` has to work around it. Align these two.

packages/data-core/src/movici_data_core/domain_model.py
```diff
@@ -0,0 +1,155 @@
...
+    config: dict = dataclasses.field(default_factory=dict)
+    references: list[MoviciDataRefInfo] = dataclasses.field(default_factory=list, compare=False)
+
+
+@dataclasses.dataclass
+class Dataset:
+    name: str
+    display_name: str
+    dataset_type: DatasetType
+    id: UUID | None = None
```
onlyname0, line 162: Same as Scenario — add `dataclasses.field(compare=False, default=None)` so equality isn't id-sensitive.

packages/data-core/src/movici_data_core/domain_model.py
```diff
@@ -0,0 +1,155 @@
...
+    workspace: Workspace | None = None
+
+    general: dict | None = None
+    espg_code: int | None = None
```
onlyname0: Typo — should be `epsg_code` (matches `Scenario.epsg_code` and the SQL column). See also `model.py:168`.

packages/data-core/src/movici_data_core/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+    pass
+
+
+class ResourceAlreadyExists(InvalidResource):
+    pass
+
+
+class InvalidAction(MoviciDataError):
+    def __init__(self, message: str | None = None):
+        self.message = message
```
onlyname0, line 183: Missing `super().__init__(message)`. Without it, `exc.args` is empty and `str(exc)` returns `""`, which will bite anyone logging or formatting this exception.

```python
def __init__(self, message: str | None = None):
    super().__init__(message)
    self.message = message
```

packages/data-core/tests/database/test_repository.py
```diff
@@ -0,0 +1,888 @@
...
+
+    async def test_list_workspaces_with_counts(
+        self, repository: SQLAlchemyRepository, filled_workspace
+    ):
+        result = await repository.workspaces.list()
+        assert len(result) == 1
+        assert result[0].dataset_count == 2
+        assert result[0].scenario_count == 1
+
+    async def filled_workspace_has_counts(self, filled_workspace):
```
onlyname0, line 60: Missing `test_` prefix — pytest never collects this. Rename to `test_filled_workspace_has_counts`.

packages/data-core/tests/database/test_backend.py
```diff
@@ -0,0 +1,177 @@
...
+        ),
+        (
+            DatabaseMode.MULTIPLE_WORKSPACES,
+            {
+                "STRICT_ATTRIBUTES": True,
+                "STRICT_DATASET_TYPES": True,
+                "STRICT_ENTITY_TYPES": True,
+                "STRICT_MODEL_TYPES": True,
+            },
+        ),
```
onlyname0, line 68: The three `flags` dicts don't include `STRICT_SCENARIO_DATASETS`, even though `_default_flags` in `general.py` sets it. Add it to all three cases (False, False, True).

packages/data-core/README.rst
```diff
@@ -0,0 +1,16 @@
+Movici Data Core
+======================
+
+Copyright 2026ff NGinfra
+
+Movici is a set of tools and software for performing simulations geospatial entities.
+
+Movici Data Core is a supporting library that holds the data and domain model for a Movici
+database. It is developed SQLAlchemy.
```
onlyname0, line 9: Missing word — "developed **with** SQLAlchemy".

PR #124, opened 2026-04-16
Generate API docs for all packages

onlyname0:
## Summary
- Adapt `docs/Makefile` apidoc target to auto-discover all packages under `packages/*/src/` using a wildcard, so new packages are included automatically
- Generate combined `modules.rst` toctree listing all discovered packages
- Fix unclosed inline literal in `WaterTankEntity` docstring

## Test plan
- [ ] Run `make apidoc` in `docs/` and verify all packages are included
- [ ] Run `make html` in `docs/` and verify the build succeeds without warnings


docs/Makefile
```diff
@@ -14,9 +14,14 @@ help:
...
+PACKAGES = $(wildcard ../packages/*/src/*)
+
 apidoc:
-	rm source/reference/*
-	sphinx-apidoc -d 1 --remove-old -t source/_templates/ -o source/reference/ ../movici_simulation_core
+	rm -f source/reference/*
+	$(foreach pkg,$(PACKAGES),\
+		sphinx-apidoc -d 1 -t source/_templates/ -o source/reference/ $(pkg) &&) true
+	printf "API Reference\n=============\n\n.. toctree::\n   :maxdepth: 1\n\n" > source/reference/modules.rst
+	$(foreach pkg,$(PACKAGES),echo "   $(notdir $(pkg))" >> source/reference/modules.rst;)
```
elfjes, line 25: I'm half inclined that we should rewrite this as a python script that uses jinja2 to render the `source/_templates/toc.rst_t` template. This would be a more readable and maintainable solution. However, we won't touch it that often (possibly never), so this ad hoc inline file generation should be fine.

Unless we start adding non-python packages to the repository (movici-flow-lib anyone?), but hey, what are the odds of that?

elfjes (approved):
Nice :)

I'm not sure about the Make shenanigans to pull this off, but it seems to do the trick. We also have the option of using ``sphinx.ext.apidoc` (https://www.sphinx-doc.org/en/master/usage/extensions/apidoc.html) to generate docs for multiple packages. But after playing a bit with it, that does not seem to make it easier to keep the output consistent with what we have now. So... 🤷

PR #128, opened 2026-05-06
Move source-type registry to SourcesSetup with lazy plugin loading

onlyname0:
## Summary
- Moves the dataset creator's source-type registry off module globals onto `SourcesSetup`, with a `register()` classmethod and a class-level `_source_types` dict.
- Adds a `movici.dataset_creator.source` entry-point group; the built-in `"file"` and `"netcdf"` types are now declared as entry points alongside any third-party types, and resolved per-name on first use.
- Drops top-level `import geopandas` / `import netCDF4` from `data_sources.py` — those backends now load only when a config actually references the corresponding source type.

## Why
This came out of review on #112: tying the registry to a class instead of a module global, picking a more descriptive entry-point group name (`movici.dataset_creator.source`), and registering built-ins through the same plugin path as third-party sources. Splitting it out keeps #112 focused on adding the EPANET source — once this lands, that PR rebases on top and uses the new group.

The lazy-import change is a related concern from the same review: a package shouldn't be imported until a config actually needs that file type. Importing `movici_simulation_core.preprocessing` (or any of its public classes) no longer pulls `geopandas` or `netCDF4` into `sys.modules`. The same pattern will apply to `wntr` once #112 rebases.

## Notes
- `register_source_type(name, cls)` is kept as a module-level shim that delegates to `SourcesSetup.register`, and is re-exported from `movici_simulation_core.preprocessing`.
- Out of scope here: integrating with the existing `Extensible` / `movici.plugins` machinery, or adding a `use_global_plugins` flag on `DatasetCreator`. Worth a follow-up.

## Test plan
- [x] `tests/preprocessing/` — 102 pass (covers existing `"file"`/`"netcdf"` paths via the new entry-point lookup).
- [x] Full simulation-core suite — 1397 pass, 1 xfail, 0 fail.
- [x] New tests: `test_unknown_source_type_raises`, `test_register_adds_source_type`.
- [x] Smoke check: `import movici_simulation_core.preprocessing` and `from … import GeopandasSource, NetCDFGridSource` both leave `sys.modules` clean of `geopandas` and `netCDF4`.


packages/simulation-core/src/movici_simulation_core/preprocessing/dataset_creator.py
```diff
@@ -132,14 +154,20 @@ def make_source(self, source_info):
...
+    @classmethod
+    def _resolve_source_type(cls, name: str) -> t.Type[DataSource]:
+        if name in cls._source_types:
+            return cls._source_types[name]
+        for entry_point in entry_points(group=SOURCE_TYPE_ENTRY_POINT_GROUP):
+            if entry_point.name == name:
+                source_cls = entry_point.load()
+                cls._source_types[name] = source_cls
+                return source_cls
+        raise ValueError(f"Unknown source type '{name}'")
```
elfjes: entry points accept a `name` filter keyword (see https://docs.python.org/3/library/importlib.metadata.html#entry-points) so that you don't need to iterate over all entry points
```suggestion
        try:
            (source_cls,) = importlib.metadata.entry_points(group=SOURCE_TYPE_ENTRY_POINT_GROUP, name=name)
            cls._source_types[name] = source_cls
            return source_cls
        except ValueError:
            raise ValueError(f"Unknown source type '{name}'") from None
```

packages/simulation-core/src/movici_simulation_core/preprocessing/dataset_creator.py
```diff
@@ -113,9 +116,28 @@ def __call__(self, dataset: dict, sources: SourcesDict) -> dict:
...
+
+    @classmethod
+    def register(cls, name: str, source_cls: t.Type[DataSource]) -> None:
+        """Register a ``DataSource`` class under ``name``.
+
+        The class must expose a ``from_source_info(source_info)`` classmethod. Source types
+        can also be registered declaratively through the
+        ``movici.dataset_creator.source`` entry-point group in a package's
+        ``pyproject.toml``.
+        """
```
elfjes, line 139: Please mention that this will override any declaratively registered source types

packages/simulation-core/tests/preprocessing/test_dataset_creator.py
```diff
@@ -389,6 +389,44 @@ def test_populates_sources_dict(self, config):
...
+
+            @classmethod
+            def from_source_info(cls, source_info):
+                cls.seen.append(source_info)
+                return cls()
+
+            def __len__(self):
+                return 0
+
+        previous = SourcesSetup._source_types.pop("dummy-test", None)
```
elfjes: suggestion, you could move the registry cleanup to a (autouse) fixture

```python
@pytest.fixture(autouse=True)
def clean_sources_setup(self):
    current_registry = {k:v for k,v in SourcesSetup._source_types.items()}
    yield
    SourcesSetup._source_types = current_registry
```

packages/simulation-core/pyproject.toml
```diff
@@ -48,12 +48,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-models = [
-  "shapely>=1.7.1",
-  "pyproj>=3.0.1",
-  "movici-geo-query>=1.1.1",
-  "scipy",
-]
+models = ["shapely>=1.7.1", "pyproj>=3.0.1", "movici-geo-query>=1.1.1", "scipy"]
```
elfjes, line 51: why the restyle?

elfjes (changes requested):
Looks good overall :). few small suggestions for improvement

packages/simulation-core/pyproject.toml
onlyname0: So apparently taplo formats short arrays inline

elfjes (approved):

PR #130, opened 2026-05-18
Prepare for movici-data-core package

elfjes:
Describe your changes
--------------------
The upcoming movici-data-core package requires changes to movici-simulation-core. Some of these changes are presented as a separted PR here

* Improve some type hints
* Add method for calculating a min and max value for specific data types
* Simplify `MoviciTypeLookup` and `MoviciDataRefInfo` to be nicer typed

Copyright license
-----------------

### DIFF
--- a/packages/simulation-core/src/movici_simulation_core/core/data_type.py
+++ b/packages/simulation-core/src/movici_simulation_core/core/data_type.py
@@ -36,6 +36,18 @@ class DataType(t.Generic[T]):
             return compare_scalar
         return compare_array
 
+    def get_min_max(self, data: np.ndarray) -> tuple[T | None, T | None]:
+        data = np.asarray(data)
+        if self.py_type is str:
+            return None, None
+        undefineds = self.is_undefined(data)
+        defined = data[~undefineds]
+        if len(defined) == 0:
+            return None, None
+        if self.py_type is bool:
+            return False, True  # type: ignore
+        return np.nanmin(defined).item(), np.nanmax(defined).item()
+
 
 UNDEFINED = {
     bool: np.iinfo(np.dtype("<i1")).min,
--- a/packages/simulation-core/src/movici_simulation_core/core/data_type.py
+++ b/packages/simulation-core/src/movici_simulation_core/core/data_type.py
@@ -44,8 +44,6 @@ class DataType(t.Generic[T]):
         defined = data[~undefineds]
         if len(defined) == 0:
             return None, None
-        if self.py_type is bool:
-            return False, True  # type: ignore
         return np.nanmin(defined).item(), np.nanmax(defined).item()
 
 
--- a/packages/simulation-core/src/movici_simulation_core/validate.py
+++ b/packages/simulation-core/src/movici_simulation_core/validate.py
@@ -10,6 +10,8 @@ from jsonschema import exceptions, validators
 
 import movici_simulation_core.core.schema as schema
 
+MoviciTypeStr = t.Literal["dataset", "entityGroup", "attribute"]
+
 
 @dataclasses.dataclass
 class ModelConfigSchema:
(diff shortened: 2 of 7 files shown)


packages/simulation-core/src/movici_simulation_core/validate.py
```diff
@@ -250,9 +254,18 @@ def movici_validator(schema, lookup: MoviciTypeLookup | None = None):
 
 @dataclasses.dataclass
 class MoviciDataRefInfo:
-    path: t.Tuple[t.Union[str, int], ...]
-    movici_type: t.Literal["dataset", "entityGroup", "attribute"]
-    value: str
+    path: tuple[str | int, ...]
```
onlyname0, line 268: Heads up: positional arg order changed (`path, movici_type, value` → `path, value, movici_type`). Tests moved to `from_path_string`, but external positional callers will silently swap `value` and `movici_type` with no error. Mind making `value` keyword-only so old calls break loudly?

packages/simulation-core/src/movici_simulation_core/core/data_type.py
```diff
@@ -36,6 +36,18 @@ def get_comparator(self, rtol=1e-5, atol=1e-8, equal_nan=False, to_scalar=False)
...
+    def get_min_max(self, data: np.ndarray) -> tuple[T | None, T | None]:
+        data = np.asarray(data)
+        if self.py_type is str:
+            return None, None
+        undefineds = self.is_undefined(data)
+        defined = data[~undefineds]
+        if len(defined) == 0:
+            return None, None
+        if self.py_type is bool:
+            return False, True  # type: ignore
```
onlyname0: This returns `(False, True)` even for `[False]` alone — is that intentional ("range of the type") or should it be the actual data min/max like the numeric path? Worth a one-line docstring either way.

packages/simulation-core/src/movici_simulation_core/validate.py
```diff
@@ -183,27 +171,43 @@ def dataset_type(self, dataset_name, required_type):
...
         self.schema = schema
 
-    def dataset(self, dataset_name):
-        return self.dataset_names is None or dataset_name in self.dataset_names
-
-    def entity_group(self, entity_type) -> bool:  # noqa: ARG002
-        return True
-
     def attribute(self, attribute_type):
         return self.schema is None or attribute_type in self.schema.attributes
```
onlyname0, line 209: I think this is a regression: with `dataset_names=["foo"]`, `super().__init__` stores `self.datasets = {"foo": None}`. The inherited `dataset_type("foo", some_type)` then does `None == some_type` → `False`, so every `movici.datasetType` check on a known dataset now fails — where the old explicit override returned `True` unconditionally. Worth restoring the override (or storing the actual types).

elfjes: The default implementation on line 165 checks if the value is either the requested type, or `None`, so `None` always passes a `movici.datasetType` check. So, no regression here. I've added some tests to show behaviour

packages/simulation-core/src/movici_simulation_core/core/data_type.py
elfjes: You're right. I initially made this choice because the intended usage is for movici-flow, which in case of a boolean always uses [False, True] for the range. However, that should be a concern for movici-flow, not here. It's better to be pure here

packages/simulation-core/src/movici_simulation_core/validate.py
elfjes: This class is purely meant as an output class, and is not meant to be instantiated by external code. The exact parameter order is therefore not part of any contract. I've added a doc string to highlight this

onlyname0 (approved):
lgtm

PR #129, opened 2026-05-18
UV: always sync all packages

elfjes:
Describe your changes
--------------------

When running uv sync for development, we should always include all packages to ensure that we install the correct (dev) dependencies

Copyright license
-----------------

Please tick the following box:
 - [x] I hereby grant a full copyright license to NGinfra to use my contribution. This includes but
 is not limited to permission to use my contribution in a commercial setting and to
 re-/sublicense my contribution

PR #132, opened 2026-05-18
Introduce movici-data-core and implement Repositories

elfjes:
## Describe your changes
Adds movici-data-core as a package and introduces the main database layer written in SQLAlchemy.

This PR depends on #130
CI depends on #129 

### Details
* the package is called `movici-data-core` and resides in `packages/data-core`
* The database model is based on modern SQLAlchemy and is async. The choice for async was driven by the need for having an async database interface in the Movici Cloud Platform (data-engine)
* A domain model exists in `domain_model.py`, which the api-contract for the SQLAlchemyRepository adheres to. SQLAlchemy models do not leak from the database layer. 
* The envisioned api layer is based on fastapi, which is fully async compatible
* A `SQLAlchemyServer` class is introduced that can setup a database, and is a factory for `SQLAlchemyBackend` classes. This backend class will be the main entry point for the api layer and is similar to a Unit of Work pattern, This class will be further implemented in a future PR
* This PR focuses on the SQLAlchemyRepository classes, which deal with reading from and writing to the database.

### Additional comments
* Initially, the package was designed so that the sqlalchemy based backend/repository could be swapped out for a file based backend, which could be used by `movici-viewer`. However, for complexity and performance reasons it was decided that movici-viewer would make use of the sqlalchemy backend, and introduce logic to import movici data files from disk and convert them to a (sqlite) database as a command.

## Copyright license

### DIFF
--- a/packages/data-core/src/movici_data_core/database/repository/dataset.py
+++ b/packages/data-core/src/movici_data_core/database/repository/dataset.py
@@ -71,6 +71,12 @@ class DatasetRepository(SQLResourceRepository):
         ).options(joinedload(db.Dataset.workspace), joinedload(db.Dataset.dataset_type))
 
     async def exists(self, name: str):
+        """return whether the dataset (by name) already exists in the database for the acitve
+        workspace.
+
+        :param name: dataset name
+        :return: a boolean indicating the dataset exists for the active workspace
+        """
         workspace_id = self._ensure_workspace_id()
         return await self._exists(db.Dataset.workspace_id == workspace_id, db.Dataset.name == name)
 
@@ -84,7 +90,16 @@ class DatasetRepository(SQLResourceRepository):
             ds.to_domain(has_data_raw, has_attributes) for ds, has_data_raw, has_attributes in rows
         ]
 
-    async def get_one_with_has_data(self, where_clause: ColumnElement[bool]):
+    async def get_by_name(self, name: str) -> Dataset | None:
+        workspace_id = self._ensure_workspace_id()
+        return await self._get_one_with_has_data(
+            (db.Dataset.workspace_id == workspace_id) & (db.Dataset.name == name)
+        )
+
+    async def get_by_id(self, id: UUID) -> Dataset | None:
+        return await self._get_one_with_has_data(db.Dataset.id == id)
+
+    async def _get_one_with_has_data(self, where_clause: ColumnElement[bool]):
         result = (
             await self.session.execute(self.selector_with_has_data.where(where_clause).limit(1))
         ).first()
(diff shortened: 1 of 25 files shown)


packages/data-core/src/movici_data_core/database/repository/updates.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        return update_id
+
+    async def delete_all(self):
+        """Delete all updates for the active scenario. Should be combined with a separate call, to
+        the Scenario, resetting the ScenarioStatus
+        """
+
+        scenario_id = self._ensure_scenario_id()
+        await self.session.execute(delete(db.Update).where(db.Update.scenario_id == scenario_id))
```
onlyname0, line 144: I think this leaks rows. CASCADE drops `UpdateAttribute` when `Update` goes, but the underlying `Attribute` (and its `DataArray` / `RowptrArray`) have no FK back — they stay. Compare with `DatasetDataRepository.delete` which explicitly walks `DatasetAttribute → Attribute`. After a few scenario resets the orphan rows will accumulate.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -0,0 +1,287 @@
...
+                [
+                    {"scenario_id": scenario_id, "dataset_id": ds.id, "sequence": idx}
+                    for idx, ds in enumerate(scenario_datasets)
+                ],
+            )
+        if not obj.models:
+            return
+
+        model_types = await self.all_data.model_types.ensure_model_types(
+            [model["type"] for model in obj.models if "name" in model]
```
onlyname0: The `if "name" in model` filter is misleading — `validators.parse_and_validate_model_config` already raises if `name` is missing, so this branch can never silently drop a model *today*. But if the validator ever loosens, this filter will produce a length-mismatched `zip` on line 225 with no error. Either drop the filter, or assert all models have a name here.

packages/data-core/src/movici_data_core/database/repository/dataset.py
```diff
@@ -0,0 +1,414 @@
...
+        """
+        dataset = await self.get_by_id(id)
+        if dataset is None:
+            raise ResourceDoesNotExist("dataset", id=id)
+
+        entity_groups: dict[str, EntityGroupSummary] = {}
+        attribute: db.Attribute
+        for attribute, min_val, max_val in await self.session.execute(
+            select(db.Attribute, db.DataArray.min_val, db.DataArray.max_val)
+            .join(db.Attribute)
```
onlyname0: The `.join(db.Attribute)` after `select(db.Attribute, db.DataArray.min_val, ...)` is hard to read — SQLAlchemy is auto-resolving it via foreign-key inference on `DataArray`, but it looks like joining `Attribute` to itself. Could you make it explicit: `.join(db.DataArray, db.DataArray.attribute_id == db.Attribute.id).join(db.DatasetAttribute, ...)`? Easier to audit and won't break if a second foreign key gets added later.

packages/data-core/src/movici_data_core/database/repository/updates.py
elfjes: you're right, thanks for the catch!

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: Yes, I meant to say `if "type" in model`, the current line looks really weird.

also, nice find on the length mismatch. I've added an assertion (with comment) and a test to make sure the assert gets tested

packages/data-core/src/movici_data_core/database/repository/dataset.py
elfjes: Make sense :)

onlyname0 (approved):

PR #135, opened 2026-05-26
Make FSM composable so that we can customize for specific use cases

elfjes:
Describe your changes
--------------------
In order to optimize the Orchestrator for a non-distributed runner (and potentially other run modes), we need to hook into the internals of the orchestrator FSMs. Currently the Orchestrator and ConnectedModel FSMs are hard coded, making it hard to change which states connect to each other. This PR decouples the transitions from the state classes and allows the FSM to be configured using an `FSMConfig` class that contains the states and the transition (conditions) between them.

Furthermore this PR

* improves typing
* refactors/simplifies/makes things more logical

Copyright license
-----------------

### DIFF
--- a/packages/simulation-core/src/movici_simulation_core/exceptions.py
+++ b/packages/simulation-core/src/movici_simulation_core/exceptions.py
@@ -14,7 +14,7 @@ class OrchestratorException(SimulationException):
     pass
 
 
-class InvalidCommand(SimulationException):
+class InvalidCommand(OrchestratorException):
     pass
 
 
--- a/packages/simulation-core/src/movici_simulation_core/exceptions.py
+++ b/packages/simulation-core/src/movici_simulation_core/exceptions.py
@@ -14,10 +14,6 @@ class OrchestratorException(SimulationException):
     pass
 
 
-class SimulationExit(OrchestratorException):
-    pass
-
-
 class InvalidCommand(SimulationException):
     pass
 
--- a/packages/simulation-core/src/movici_simulation_core/exceptions.py
+++ b/packages/simulation-core/src/movici_simulation_core/exceptions.py
@@ -18,6 +18,10 @@ class SimulationExit(OrchestratorException):
     pass
 
 
+class InvalidCommand(SimulationException):
+    pass
+
+
 class NotReady(SimulationException):
     pass
 
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -278,16 +278,18 @@ class WaitingForMessage(BaseModelState):
 
     @_handle_response.register
     def _(self, msg: AcknowledgeMessage) -> None:
-        """when a model sends an accepted message, don't do extra logic"""
+        """when a model sends an acknowledge message, do nothing but notify subscribers that the
+        model has returned
+        """
         if not self.context.pending_updates:
-            self.notify_subscribers()
+            self.notify_subscribers(NoUpdateMessage())
 
     @_handle_response.register
     def _(self, msg: ResultMessage) -> None:
-        """When a result comes set the model's next_time and possibly add a message to the
-        subscribers queues"""
+        """When a result comes set the model's next_time and notify subscribers that this model
+        has returned, with possible data"""
         self.context.timeline.set_next_time(self.context, msg.next_time)
-        command = None
+        command = NoUpdateMessage()
         if msg.has_data:
             command = UpdateMessage(
                 timestamp=self.context.timeline.current_time,
@@ -311,9 +313,14 @@ class WaitingForMessage(BaseModelState):
             self.context.pending_quit = QuitMessage()
             self.context.pending_updates = []
 
-    def notify_subscribers(self, command: Command | None = None):
-        command = command or NoUpdateMessage()
+    def notify_subscribers(self, command: UpdateMessage | NoUpdateMessage):
+        """Notify subscribers that this ConnectedModel's model has returned."""
         for model in self.context.publishes_to:
+            # This method may technically raise InvalidCommand if any of the subscribed models
+            # does not accept an UpdateMessage or NoUpdateMessage, but the (default)
+            # ``ConnectedModel.fsm_config`` ensures that this cannot happen. In case of a custom
+            # ``fsm_config`` and this method raises ``InvalidCommand``, this is caught upstream by
+            # the orchestrator's ``Config``
             model.recv_event(command)
 
 
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass, field
 from functools import singledispatchmethod
 from itertools import product
 
+from movici_simulation_core.exceptions import InvalidCommand
 from movici_simulation_core.messages import (
     AcknowledgeMessage,
     ErrorMessage,
@@ -32,6 +33,7 @@ class Context:
     global_timer: Stopwatch = field(default=None)
     phase_timer: Stopwatch = field(default=None)
     logger: logging.Logger = field(default_factory=logging.getLogger)
+    orchestrator_failed: bool = False
 
     def __post_init__(self):
         self.global_timer = self.global_timer or ReportingStopwatch(
@@ -43,7 +45,7 @@ class Context:
 
     @property
     def failed(self):
-        return self.models.failed
+        return self.orchestrator_failed or self.models.failed
 
     def log_new_phase(self, phase: str):
         self.logger.info(f"Entering {phase}")
@@ -58,11 +60,23 @@ class Context:
 
         if next_time != self.timeline.current_time:
             self.timeline.current_time = next_time
-            self.models.queue_all(NewTimeMessage(next_time))
+            self.recv_for_all(NewTimeMessage(next_time))
 
         for model in self.models.values():
             if model.next_time == next_time:
-                model.recv_event(UpdateMessage(timestamp=next_time))
+                self.recv_message(model, UpdateMessage(timestamp=next_time))
+
+    def recv_for_all(self, message: Message):
+        """Receive a Message for all models"""
+        for model in self.models.values():
+            self.recv_message(model, message)
+
+    def recv_message(self, model: ConnectedModel, message: Message):
+        try:
+            model.recv_event(message)
+        except InvalidCommand:
+            self.logger.exception("Error while sending command")
+            self.orchestrator_failed = True
 
     def log_new_time(self):
         if self.models.next_time != self.timeline.current_time:
@@ -213,7 +227,7 @@ class WaitingForMessage(BaseModelState):
 
     def process_command(self, msg: Command):
         if not isinstance(msg, self.valid_commands):
-            raise ValueError(
+            raise InvalidCommand(
                 f"Received invalid command {msg} for model '{self.context.name}'. Expected one of "
                 + ", ".join(m.__name__ for m in self.valid_commands)
             )
@@ -430,11 +444,6 @@ class ModelCollection(dict[bytes, ConnectedModel]):
     @property
     def failed(self):
         return [model.name for model in self.values() if model.failed]
-
-    def queue_all(self, message: Message):
-        """add a message to the queue of all models"""
-        for model in self.values():
-            model.recv_event(message)
 
     def determine_interdependency(self):
         """calculate the subscribers for every model based on the pub/sub mask."""
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -155,11 +155,10 @@ class ConnectedModel:
     busy: bool = field(default=False, init=False)
     next_time: t.Optional[int] = field(default=None, init=False)
     failed: bool = field(default=False, init=False)
-    ack: bool = field(default=False, init=False)
 
-    quit: t.Optional[QuitMessage] = field(default=None, init=False)
+    pending_new_time: NewTimeMessage | None = field(default=None, init=False)
     pending_updates: t.List[UpdateMessage] = field(default_factory=list, init=False)
-    processing_new_time: bool = False
+    pending_quit: QuitMessage | None = field(default=None, init=False)
 
     fsm_config: FSMConfig | None = None
     fsm: FSM[ConnectedModel, Message] = field(init=False)
@@ -187,7 +186,7 @@ class ConnectedModel:
 
     def log_invalid(self, message, valid_messages: t.Iterable[t.Type[Message]]):
         self.logger.error(
-            f"Received invalid message in {message} from model '{self.name}'. Expected one of "
+            f"Received invalid message {message} from model '{self.name}'. Expected one of "
             + ", ".join(m.__name__ for m in valid_messages)
         )
 
@@ -239,16 +238,16 @@ class WaitingForMessage(BaseModelState):
             self._handle_response(msg)
 
     def process_new_time(self, msg: NewTimeMessage):
-        pass
+        self.context.pending_new_time = msg
 
     def process_no_update(self, msg: NoUpdateMessage):
         pass
 
     def process_update(self, msg: UpdateMessage):
-        pass
+        self.context.pending_updates.append(msg)
 
     def process_quit(self, msg: QuitMessage):
-        pass
+        self.context.pending_quit = msg
 
     @singledispatchmethod
     def _handle_response(self, msg: Message) -> None:
@@ -263,7 +262,6 @@ class WaitingForMessage(BaseModelState):
     @_handle_response.register
     def _(self, msg: AcknowledgeMessage) -> None:
         """when a model sends an accepted message, don't do extra logic"""
-        self.context.ack = True
         if not self.context.pending_updates:
             self.notify_subscribers()
 
@@ -286,14 +284,14 @@ class WaitingForMessage(BaseModelState):
     def _(self, msg: ErrorMessage) -> None:
         """When a model reports an error, set it to failed"""
         self.context.failed = True
-        self.context.quit = None
+        self.context.pending_quit = None
         self.context.pending_updates = []
 
-    def handle_invalid_response(self, msg: Message, valid_messages: t.Iterable[t.Type[Message]]):
+    def handle_invalid_response(self, msg: Response, valid_messages: t.Iterable[t.Type[Response]]):
         self.context.failed = True
         self.context.log_invalid(msg, valid_messages)
-        if not self.context.quit:
-            self.context.quit = QuitMessage()
+        if not self.context.pending_quit:
+            self.context.pending_quit = QuitMessage()
             self.context.pending_updates = []
 
     def notify_subscribers(self, command: t.Optional[Command] = None):
@@ -307,17 +305,6 @@ class Idle(WaitingForMessage):
 
     valid_commands = (NewTimeMessage, NoUpdateMessage, UpdateMessage, QuitMessage)
     valid_responses = ()
-
-    def process_new_time(self, msg: NewTimeMessage):
-        self.context.ack = False
-        self.context.processing_new_time = True
-        self.context.send_command(msg)
-
-    def process_update(self, msg: UpdateMessage):
-        self.context.pending_updates.append(msg)
-
-    def process_quit(self, msg: QuitMessage):
-        self.context.quit = msg
 
 
 class Busy(WaitingForMessage):
@@ -333,12 +320,6 @@ class Busy(WaitingForMessage):
         self.context.busy = True
         self.context.timer.start()
 
-    def process_update(self, msg: UpdateMessage):
-        self.context.pending_updates.append(msg)
-
-    def process_quit(self, msg: QuitMessage):
-        self.context.quit = msg
-
 
 class Registration(Busy):
     """A RegistrationMessage is expected from the model"""
@@ -348,14 +329,17 @@ class Registration(Busy):
 
 
 class NewTime(Busy):
-    """A NewTime message has been sent to the model, and it needs to Acknowledge that"""
+    """A NewTime message will be sent to the model, and it needs to Acknowledge that"""
 
     valid_commands = (UpdateMessage, NoUpdateMessage, QuitMessage)
     valid_responses = (AcknowledgeMessage, ErrorMessage)
 
     def on_enter(self):
+        if not isinstance(self.context.pending_new_time, NewTimeMessage):
+            raise RuntimeError("can only enter NewTime when there is a pending NewTimeMessage")
+        self.context.send_command(self.context.pending_new_time)
+        self.context.pending_new_time = None
         super().on_enter()
-        self.context.processing_new_time = False
 
 
 class Updating(Busy):
@@ -372,9 +356,6 @@ class PendingMoreUpdates(Idle):
 
     valid_commands = (NoUpdateMessage, UpdateMessage, QuitMessage)
     valid_responses = ()
-
-    def process_update(self, msg: UpdateMessage):
-        self.context.pending_updates.append(msg)
 
 
 class ProcessPendingUpdates(BaseModelState):
@@ -409,12 +390,11 @@ class ProcessPendingQuit(BaseModelState):
     """
 
     def run(self):
-        if not isinstance(self.context.quit, QuitMessage):
+        if not isinstance(self.context.pending_quit, QuitMessage):
             raise RuntimeError(
                 "can only enter ProcessPendingQuit when there is a QuitMessage pending"
             )
-        self.context.ack = False
-        self.context.send_command(self.context.quit)
+        self.context.send_command(self.context.pending_quit)
 
 
 class Finalizing(Busy):
@@ -439,10 +419,6 @@ class ModelCollection(dict[bytes, ConnectedModel]):
     @property
     def busy(self):
         return any(model.busy for model in self.values())
-
-    @property
-    def waiting_for(self):
-        return [model for model in self.values() if model.busy]
 
     @property
     def next_time(self):
@@ -479,7 +455,7 @@ class ModelHasFailed(Condition[ConnectedModel]):
 
 class HasPendingQuitAndIdle(Condition[ConnectedModel]):
     def met(self) -> bool:
-        return bool(self.context.quit and not self.context.busy)
+        return bool(self.context.pending_quit and not self.context.busy)
 
 
 class HasPendingUpdatesAndIdle(Condition[ConnectedModel]):
@@ -487,9 +463,9 @@ class HasPendingUpdatesAndIdle(Condition[ConnectedModel]):
         return bool(self.context.pending_updates and not self.context.busy)
 
 
-class IsProcessingNewTime(Condition[ConnectedModel]):
+class HasPendingNewTimeAndIdle(Condition[ConnectedModel]):
     def met(self) -> bool:
-        return self.context.processing_new_time
+        return bool(self.context.pending_new_time)
 
 
 class IsIdle(Condition[ConnectedModel]):
@@ -518,7 +494,7 @@ MODEL_FSM_CONFIG = FSMConfig(
         Idle: [
             (HasPendingQuitAndIdle, ProcessPendingQuit),
             (HasPendingUpdatesAndIdle, ProcessPendingUpdates),
-            (IsProcessingNewTime, NewTime),
+            (HasPendingNewTimeAndIdle, NewTime),
         ],
         ProcessPendingUpdates: [
             (ModelWaitingForDependencies, PendingMoreUpdates),
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -465,7 +465,7 @@ class HasPendingUpdatesAndIdle(Condition[ConnectedModel]):
 
 class HasPendingNewTimeAndIdle(Condition[ConnectedModel]):
     def met(self) -> bool:
-        return bool(self.context.pending_new_time)
+        return bool(self.context.pending_new_time and not self.context.busy)
 
 
 class IsIdle(Condition[ConnectedModel]):
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -221,9 +221,9 @@ class WaitingForMessage(BaseModelState):
     def recv_message(self, msg: Message):
         # Here we make use of the fact that for a typing.Union object, __args__ contains the
         # classes inside the union
-        if isinstance(msg, Command.__args__):
+        if isinstance(msg, Command):
             self.process_command(msg)
-        elif isinstance(msg, Response.__args__):
+        elif isinstance(msg, Response):
             self.handle_response(msg)
         else:
             raise TypeError("Unknown Message")
@@ -320,7 +320,7 @@ class WaitingForMessage(BaseModelState):
             # does not accept an UpdateMessage or NoUpdateMessage, but the (default)
             # ``ConnectedModel.fsm_config`` ensures that this cannot happen. In case of a custom
             # ``fsm_config`` and this method raises ``InvalidCommand``, this is caught upstream by
-            # the orchestrator's ``Config``
+            # the orchestrator's ``Context``
             model.recv_event(command)
 
 
@@ -328,6 +328,14 @@ class Idle(WaitingForMessage):
     """Base class for if a model is awaiting further instructions"""
 
     valid_commands = (NewTimeMessage, NoUpdateMessage, UpdateMessage, QuitMessage)
+
+    # In an Idle state, we don't expect any messages from a model. It is supposed to wait for
+    # further instructions. The model's current Stream implementation blocks the zmq.Socket while
+    # idle, so there is no way that the model can actually send a message. However, future
+    # implementations may allow for a model to asynchronously send a message that is not a response
+    # to a command, most likely to send an ErrorMessage if it has crashed due to external
+    # circumstance. Until that time tough, there are no valid responses while idle, since there is
+    # nothing to repond to
     valid_responses = ()
 
 
--- a/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
+++ b/packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
@@ -20,7 +20,7 @@ from movici_simulation_core.messages import (
 )
 from movici_simulation_core.utils.data_mask import masks_overlap
 
-from .fsm import FSM, Always, State, TransitionsT
+from .fsm import FSM, Always, Condition, FSMConfig, State
 from .interconnectivity import Publisher, format_matrix
 from .stopwatch import ReportingStopwatch, Stopwatch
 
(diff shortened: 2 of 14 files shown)


onlyname0 (commented):
Nice refactor — pulling the transitions out of the `State` classes into a declarative `FSMConfig` makes the orchestrator FSM much easier to reason about, and the `Publisher` protocol cleanup is a genuine fix (the old `IPubSubFilter` declared a `subscribers` attribute that `ConnectedModel` never had, while `format_matrix` was already using `publishes_to`).

I traced the new condition tables against the old hard-coded transitions and the behaviour is equivalent on all the normal paths; the full simulation-core suite + lint are green locally.

One thing I'd like to confirm before approving (see inline on `process_command`): making an invalid command a hard `ValueError` is reasonable as an invariant, but the exception isn't caught anywhere and would propagate out of the orchestrator. Since the whole point of this PR is to enable custom FSMs / run modes, that's exactly where the invariant is most likely to be violated. Happy to approve once we've settled on the intended failure mode there.

A few test-coverage gaps I noticed (all fail-fast guards, low risk, optional): the `FSMConfig.strict` validation paths, the `RuntimeError` in `Context.queue_models_for_next_time` when `next_time is None`, and the `RuntimeError` in `NewTime.on_enter` without a pending new-time — none are currently exercised.

The rest are minor / nits, inline.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -209,8 +213,10 @@ def recv_message(self, msg: Message):
 
     def process_command(self, msg: Command):
         if not isinstance(msg, self.valid_commands):
-            self.handle_invalid(msg, self.valid_commands)
-            return
+            raise ValueError(
```
onlyname0: This changes an invalid command from graceful degradation (set `failed`, queue a `QuitMessage`) to a raised `ValueError`, and the raise isn't caught on the way out: `send_silent` only swallows `StopIteration`, the model and orchestrator FSM `_run` loops only catch `FSMException`, and `Orchestrator.run` only catches `(FSMDone, FSMError)` — so it propagates all the way up and takes the orchestrator down.

I see from `ccf6cdb` this is intentional ("invalid command is a hard error so that we can simplify model fsm transitions"), and it's unreachable on today's paths. My only concern is that this PR exists to make the FSM configurable for new run modes, and a custom config is precisely where a model could legitimately receive a command its state doesn't list. Could we either catch this at the orchestrator boundary and route it through the existing failure/quit handling, or leave a comment documenting that an invalid command is a hard invariant violation by design? Just want the failure mode to be deliberate.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
```diff
@@ -18,27 +20,59 @@ def send_silent(coro: t.Generator, value: t.Any):
...
+        if self.initial_state not in self.states:
+            raise ValueError(
+                f"'initial_state' {self.initial_state.__name__} is not a member of 'states' "
+                "and 'strict' was set"
+            )
+        for _, state in itertools.chain.from_iterable(self.states.values()):
+            if state not in self.states:
+                raise ValueError(
+                    f"State {state.__name__} was mentioned in a transition but it is not a member"
+                    "of 'states' and 'strict' was set"
```
onlyname0: Adjacent string literals concatenate without a space here, so this renders as `…it is not a memberof 'states'…`.

```suggestion
                    f"State {state.__name__} was mentioned in a transition but it is not a "
                    "member of 'states' and 'strict' was set"
```

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -134,32 +146,35 @@ class ConnectedModel:
...
     failed: bool = field(default=False, init=False)
-    ack: bool = field(default=False, init=False)
 
-    quit: t.Optional[QuitMessage] = field(default=None, init=False)
+    pending_new_time: NewTimeMessage | None = field(default=None, init=False)
     pending_updates: t.List[UpdateMessage] = field(default_factory=list, init=False)
+    pending_quit: QuitMessage | None = field(default=None, init=False)
 
-    fsm: FSM[ConnectedModel] = field(init=False)
+    fsm_config: FSMConfig | None = None
```
onlyname0: Since the PR tightens typing, worth parameterising this to match the `fsm` field on the next line (`FSM[ConnectedModel, Message]`):

```suggestion
    fsm_config: FSMConfig[ConnectedModel] | None = None
```

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -134,32 +146,35 @@ class ConnectedModel:
...
-    pub: t.Optional[dict] = field(default_factory=dict)
-    sub: t.Optional[dict] = field(default_factory=dict)
+    publishes_to: list[ConnectedModel] = field(default_factory=list)
+    subscribed_to: list[ConnectedModel] = field(default_factory=list)
+    timer: Stopwatch = field(init=False)
+    pub: dict | None = field(default_factory=dict)
+    sub: dict | None = field(default_factory=dict)
 
     busy: bool = field(default=False, init=False)
     next_time: t.Optional[int] = field(default=None, init=False)
```
onlyname0: Small consistency nit: this and `pending_updates: t.List[UpdateMessage]` (L160) still use legacy `t.Optional` / `t.List`, while the surrounding new fields use `… | None` / `list[…]`. Could modernise both to `int | None` and `list[UpdateMessage]` to match.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -168,8 +183,6 @@ def recv_event(self, event: Message):
     def send_command(self, message: Command) -> None:
         """Send a message and start the timer. Also, start waiting"""
```
onlyname0: Docstring is stale — the timer-start / busy logic moved to `Busy.on_enter`, so this method only sends now.

```suggestion
        """Send a command message to the model."""
```

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: good point, I've modified the logic so that it fails the orchestrator itself, and - if the orchestrator otherwise handles QuitMessages properly - will send it to all models so that the simulation can be shut down gracefully. This is similar to the previous behaviour

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
elfjes: thanks!

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: thanks!

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: thanks!

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: Thanks!

elfjes:
While working on this I (re)discovered that debugging the orchestrator was very difficult due to the coroutine based approach for the FSM. I've gone ahead and simplified the design, so that states can implement either `handle_event` or `run` depending on whether the state can be run directly or that it needs input from outside

packages/simulation-core/tests/services/orchestrator/test_states.py
```diff
@@ -56,14 +56,11 @@ def context(self, context, model_mock):
...
     def send_message(self, state, event):
-        runner = state.run()
-        send_silent(runner, None)
 
         def _send(name=None, msg=None):
             msg = msg or event
             name = name or "model_a"
-            send_silent(runner, (name, msg))
-            return runner
+            state.handle_event((name, msg))
```
onlyname0, line 62: This fixture migration (coroutine-driving → `state.handle_event`) left two downstream tests asserting on a method production never calls. `WaitForModels.handle_event` routes through `context.recv_message` → `model.recv_event` — there is no `handle_message`. So:

- **`test_quits_when_model_crashes` (L72-75)**: `model_mock.handle_message.side_effect = SimulationExit` never fires. The assertion `context.failed == ['model_a']` passes *only* because a bare `Mock`'s `.failed` is truthy by default — set `model_mock.failed = False` and it becomes `[]` and the test fails. It verifies nothing about crash handling.
- **`test_ignores_unknown_models` (L77-79)**: `handle_message.call_count == 0` is always 0 (even for a *known* model, since `recv_event` is the real call site), so it can't distinguish "unknown ignored" from "known processed".

Suggest driving the real path (e.g. `model_mock.recv_event.side_effect = …` and assert `recv_event.call_args`) and, for the unknown case, asserting `model_mock.recv_event.call_count == 0`.

packages/simulation-core/tests/services/orchestrator/test_orchestrator.py
```diff
@@ -733,3 +682,61 @@ def test_acknowledge_message_doesnt_trigger_pending_model_when_dependency_should
...
+            ("b", RegistrationMessage(pub=None, sub=None)),
+            ("b", AcknowledgeMessage()),
+        ],
+    )
+    assert results == [
+        ("b", QuitMessage(due_to_failure=True)),
+    ]
+
+
+def test_invalid_command_gracefully_shuts_down_models(run_orchestrator):
```
onlyname0, line 719: Good end-to-end coverage of the graceful-shutdown *mechanism*. One gap: this swaps in a bespoke `FailingOrchestrator` / `failing_fsm_config` whose failing `IncorrectState` already carries `(Failed, StartFinalizingPhase)`. The production `FSM_CONFIG` is never driven into `orchestrator_failed` mid-run, so the real config's failure seam isn't exercised (see the `NewTime` note on `service.py`). Consider an integration test on the real `Orchestrator` that induces an `InvalidCommand` during the running phase and asserts every still-live model receives `QuitMessage(due_to_failure=True)` and the run exits with code 1. Keep this one as the mechanism-level unit test.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/service.py
```diff
@@ -8,9 +8,53 @@
...
+        ],
+        ModelsRegistration: [
+            (Failed, StartFinalizingPhase),
+            (AllModelsReady, StartRunningPhase),
+        ],
+        StartRunningPhase: [
+            (Always, NewTime),
+        ],
+        NewTime: [
+            (Always, WaitForResults),
```
onlyname0, line 41: Consistency/defensiveness: `NewTime` has only `(Always, WaitForResults)`, whereas `ModelsRegistration` and `WaitForResults` both lead with `(Failed, StartFinalizingPhase)`. `NewTime.run()` → `queue_models_for_next_time()` → `recv_message()` can set `orchestrator_failed`, but without a `Failed` edge here that's only acted on when the *next* inbound model event arrives. In practice this is effectively unreachable (entering `NewTime`, all models are `Idle`, whose `valid_commands` accept both `NewTimeMessage` and `UpdateMessage`), so it's not a bug — but adding `(Failed, StartFinalizingPhase)` here would make orchestrator-failure handling immediate and symmetric with the other waiting states.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
```diff
@@ -11,55 +12,77 @@
...
 
+    def __init__(self, config: FSMConfig[T], context: T = None, raise_on_done=True):
 
-class FSM(t.Generic[T, E]):
-    def __init__(self, initial_state: t.Type[State], context: T = None, raise_on_done=True):
         self.context = context
-        self.state = initial_state(context=self.context)
+        self.config = config
+        self.state = config.initial_state(context=self.context)
         self.runner = None
```
onlyname0: Dead since the coroutine removal: `self.runner` is now assigned once and never read again (the former readers `send_silent`/`self.runner = self._run()`/`assert self.runner is not None` are all gone). Safe to delete.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
```diff
@@ -11,55 +12,77 @@
...
-        send_silent(self.runner, None)
-
-    def _run(self):
+        with self.catch_fsmexceptions():
+            self.state.on_enter()
+            self.run_until_event_required()
+
+    def handle_event(self, event: E):
+        self.ensure_not_done()
+        assert self.state.requires_event
```
onlyname0, line 114: `assert` is compiled out under `python -O`. The invariant (the FSM always rests on a `requires_event` state after `start()`/`handle_event()`) holds structurally today, so this never actually fires. But if it ever did under `-O`, control would fall through to the base `State.handle_event`'s `NotImplementedError` which — not being an `FSMException` — escapes `catch_fsmexceptions` and crashes the stream loop. Low priority, but raising a dedicated `FSMException` subclass instead would survive `-O` and route through the normal shutdown path.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -279,18 +301,17 @@ def _(self, msg: ResultMessage) -> None:
...
+        if not self.context.pending_quit:
+            self.context.pending_quit = QuitMessage()
             self.context.pending_updates = []
-            self.next_state = ProcessPendingQuit
 
-    def notify_subscribers(self, command: t.Optional[Command] = None):
+    def notify_subscribers(self, command: Command | None = None):
         command = command or NoUpdateMessage()
         for model in self.context.publishes_to:
             model.recv_event(command)
```
onlyname0, line 324: `notify_subscribers` is now the only model-dispatch site not routed through the new `context.recv_message` wrapper. It's harmless in practice (subscribers only receive `Update`/`NoUpdate`, which every state reachable once `publishes_to` is populated accepts; and any `InvalidCommand` would still be caught by the enclosing `recv_message` that dispatched to the publisher), but the asymmetry is easy to misread as a hole. A one-line comment explaining why this site is intentionally raw — or routing it through the wrapper for symmetry — would help future readers.

onlyname0 (commented):
## Re-review — changes since the 2026-06-02 review

Thanks for working through all the feedback. I re-reviewed the diff since my last pass (the FSM rewrite + graceful error handling).

**All five earlier comments are resolved** — `process_command` redesigned into the `InvalidCommand` graceful path, the `member`/`of` space, the `fsm_config` generic, the `t.Optional`/`t.List` modernization (no leftovers anywhere in the touched files), and the `send_command` docstring. 👍

**The coroutine→dispatch FSM rewrite checks out.** I traced it end-to-end: `on_enter` fires exactly once per state entry, `start()`/`handle_event()` mirror the old `{ run(); transition() }` loop, and `FSMDone`/`FSMError` still propagate out under `raise_on_done=True` so `Orchestrator.run()` terminates correctly. `send_silent`/`fsm.send`/`Event`/`State.transitions()` are fully migrated (no dangling refs). Clean simplification, and the graceful `InvalidCommand` handling (a `SimulationException`, deliberately *not* an `FSMException`, so the model FSM lets it through to `Context.recv_message`) is well-factored.

Inline comments below, roughly by priority:
- **(test)** the `send_message` migration in `test_states.py` left two tests asserting on a `handle_message` that production never calls — `test_quits_when_model_crashes` ends up verifying nothing. This is the one I'd most want fixed.
- **(coverage)** the graceful-shutdown e2e test uses a bespoke FSM config; the real `FSM_CONFIG` failure path isn't exercised.
- **(consistency)** `NewTime` has no `(Failed, …)` transition.
- **(cleanup)** dead `self.runner`; an `assert` that's stripped under `-O`; `notify_subscribers` bypassing `recv_message`.

A few pure nits not worth their own threads: `transition()` both mutates `self.state` and returns it while both callers do `self.state = self.transition()` (redundant — pick one style); a stray blank line right after the `FSM.__init__` signature; `State(ABC)` now has no abstract methods, so the ABC base is nominal (fine, given the run-XOR-handle_event split); and `InvalidCommand` could subclass `OrchestratorException` instead of `SimulationException` for precision (just don't move it under `FSMException` — that would make `catch_fsmexceptions` swallow it and break the design).

Posting as a comment rather than an approve — the vacuous test is the only thing I'd want addressed first; nothing here is a blocker.

packages/simulation-core/tests/services/orchestrator/test_states.py
elfjes: You're right. Those tests are currently useless. I've fixed the unknown-model case.

However, I'm struggling to find the exact code path or failure mode that was supposed to be tested by the `test_quits_when_model_crashes` case. I think an older version of the Orchestrator would raise a `SimulationExit` in case a model (on the other side of the zmq connection) did something wrong that could not be processed by the `ConnectedModel` FSM; that is now handled differently. 

Currently, `SimulationExit` is not used anymore, so I think I'll just remove that exception. The only other exception that's explicitly raised, is `InvalidCommand`, which is already tested for in other places. The `test_quits_when_model_crashes` then does not have a purpose anymore and can just go

packages/simulation-core/tests/services/orchestrator/test_orchestrator.py
elfjes: Yeah, thats the thing. There are no code paths in the current Orchestrator where it sends an invalid command; and if there were, then that is a bug that needs to be fixed. This graceful shutdown mechanism is purely there for when someone decides to compose their own Orchestrator and leaves a bug in there

packages/simulation-core/src/movici_simulation_core/services/orchestrator/service.py
elfjes: That's because `NewTime(OrchestratorState)` is a State that is not `requires_event  = True`, and there is possibility of it resulting in a `Failed` condition. Since the only way the `Failed` condition can be met, is if there was either an invalid command, or after a model sends a response. Neither can happen in the `NewTime(OrchestratorState)` (not to be confused with the `NewTime(BaseModelState)` connected model state)

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
elfjes: good catch, thanks :)

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
elfjes: This assert is there to ensure the invariant that whenever this callback is called, the current state is expecting a `State.handle_event` call. This is assured because inside every `FSM.handle_event` call, we run until the next `requires_event` State through `FSM.run_until_event_required()`

If this `assert` raises, that's due to a programmer making a mistake (which should be caught in testing) rather than an actual code path that we want do handle in runtime

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: Sure :)

onlyname0 (commented):
## Follow-up review — model-side FSM (the declarative-transition translation)

Went deeper on the part the earlier passes touched least: the model FSM. The mechanism rewrite is sound, but flattening the model FSM to a **pure declarative `MODEL_FSM_CONFIG`** dropped a capability the base relied on.

**Root cause (one issue, several symptoms):** in the base, `BaseModelState.transitions()` returned the per-event `self.next_state` override *if a handler set one*, otherwise the declarative `_transitions()`. Handlers used this to **divert** the next transition — e.g. `handle_invalid` set `next_state = ProcessPendingQuit`, and `Idle.process_quit` set `next_state = ProcessPendingQuit`. The new pure-declarative config can't express those diversions, so a few model states now resolve to the wrong successor. Details inline.

Summary of inline comments:
- 🔴 **blocker** — a model in `PendingMoreUpdates` that receives any `Response` (e.g. a crashing model's `ErrorMessage`) hits an uncaught `RuntimeError` that kills the orchestrator. The base handled this gracefully — so it's a regression, and it cuts against this PR's own graceful-error-handling goal.
- 🟠 **major** — `MODEL_BUSY_TRANSITIONS` orders quit before failed (base was failed-first) → a failed model holding a pending quit re-enters `Finalizing` (busy) instead of `Done`, risking a finalize-phase hang.
- 🟠 the same `PendingMoreUpdates` entry also ignores a `QuitMessage` during finalize.
- composability footguns + a couple of nits.

I verified the blocker and the two divergences directly against the base source (the base's two-tier `transitions()` diverts where the new config can't). One trivial body-only nit not worth its own thread: `test_states.py:67` sets `state.valid_messages = ("dummy",)` which is never read (stale).

Holding off on approve because of the blocker. Happy to draft the `MODEL_FSM_CONFIG` fix + regression tests if useful.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -491,3 +467,69 @@ def determine_interdependency(self):
...
+            (HasPendingUpdatesAndIdle, ProcessPendingUpdates),
+            (HasPendingNewTimeAndIdle, NewTime),
+        ],
+        ProcessPendingUpdates: [
+            (ModelWaitingForDependencies, PendingMoreUpdates),
+            (Always, Updating),
+        ],
+        PendingMoreUpdates: [
+            (Always, ProcessPendingUpdates),
+        ],
```
onlyname0, line 534: 🔴 **Blocker: a single model error here can crash the whole orchestrator (regression).**

`PendingMoreUpdates`'s only transition is `(Always, ProcessPendingUpdates)`. But `PendingMoreUpdates` has `valid_responses=()`, so **any** inbound `Response` (most realistically a crashing model's `ErrorMessage`) → `handle_invalid_response`, which sets `failed=True`, `pending_quit=QuitMessage()`, and **clears `pending_updates=[]`**. The forced `(Always, ProcessPendingUpdates)` then runs `ProcessPendingUpdates.run()`, which does `if not updates: raise RuntimeError(...)`. `RuntimeError` is neither `FSMException` nor `InvalidCommand`, so it escapes the model FSM's `catch_fsmexceptions`, `Context.recv_message`, and `Stream.run` → the orchestrator dies.

This is a regression: in the base, `handle_invalid` set `self.next_state = ProcessPendingQuit`, and `BaseModelState.transitions()` returned that override, diverting away from `ProcessPendingUpdates` → no crash (model went to `Finalizing`). The pure-declarative config can't express that diversion. Plain `Idle` survives the same input because its list leads with `(HasPendingQuitAndIdle, ProcessPendingQuit)`; only `PendingMoreUpdates` lacks those guards.

The same entry also causes a second issue: a `QuitMessage` arriving while a dependency is busy is *ignored* (base diverted to `ProcessPendingQuit` via `process_quit`); the model instead computes an update during the finalize phase.

**Fix:** give `PendingMoreUpdates` an `Idle`-like guarded list, e.g.
```python
PendingMoreUpdates: [
    (HasPendingQuitAndIdle, ProcessPendingQuit),
    (ModelHasFailed, Done),
    (Always, ProcessPendingUpdates),
],
```
and consider making `ProcessPendingUpdates.run()` transition gracefully instead of `raise` on empty `pending_updates` (defense-in-depth, since the precondition is now violable by external model input). Worth a regression test: model in `PendingMoreUpdates` receiving `ErrorMessage` / a stray `ResultMessage`.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -491,3 +467,69 @@ def determine_interdependency(self):
...
+    def met(self) -> bool:
+        return any(model.busy for model in self.context.subscribed_to)
+
+
+MODEL_BUSY_TRANSITIONS = (
+    (HasPendingQuitAndIdle, ProcessPendingQuit),
+    (ModelHasFailed, Done),
+    (HasPendingUpdatesAndIdle, ProcessPendingUpdates),
+    (IsIdle, Idle),
+)
```
onlyname0, line 515: 🟠 **Major: quit-vs-failed priority is inverted vs the base → finalize-phase hang risk.**

`MODEL_BUSY_TRANSITIONS` lists `(HasPendingQuitAndIdle, ProcessPendingQuit)` **before** `(ModelHasFailed, Done)`. The base `Busy._transitions()` listed `failed → Done` **first**. These differ for a Busy model that fails *while already holding a pending quit* — which is reachable during finalize:

1. Model A fails → orchestrator `WaitForResults` takes `(Failed, StartFinalizingPhase)` while model B is still busy.
2. `StartFinalizingPhase.run()` → `recv_for_all(QuitMessage)` → B gets `pending_quit` set (stays busy).
3. B then returns an invalid response → `handle_invalid_response` sets `failed=True` but leaves the existing `pending_quit` (skips the `if not pending_quit` block).

Now `failed=True, pending_quit set, busy=False`:
- **Base:** `failed` wins → `Done` (not busy, nothing sent).
- **HEAD:** `HasPendingQuitAndIdle` wins → `ProcessPendingQuit` → `Finalizing` with `busy=True`, waiting for an ack from a model that just misbehaved → `FinalizingWaitForModels` (gated on `not models.busy`) can hang.

**Fix (pick one):** put `(ModelHasFailed, Done)` before `(HasPendingQuitAndIdle, ProcessPendingQuit)` here, **or** have `handle_invalid_response` clear `pending_quit` (mirroring the `ErrorMessage` handler, which already sets `pending_quit=None`). See the related comment on `handle_invalid_response`.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -279,20 +303,24 @@ def _(self, msg: ResultMessage) -> None:
...
-    def handle_invalid(self, msg: Message, valid_messages: t.Iterable[t.Type[Message]]):
-        self.context.log_invalid(msg, valid_messages)
+    def handle_invalid_response(self, msg: Response, valid_messages: t.Iterable[type[Response]]):
         self.context.failed = True
-        if not self.context.quit:
-            self.context.quit = QuitMessage()
+        self.context.log_invalid(msg, valid_messages)
+        if not self.context.pending_quit:
+            self.context.pending_quit = QuitMessage()
             self.context.pending_updates = []
```
onlyname0, line 314: This is the other half of the two issues above. Two asymmetries to consider:

1. It clears `pending_updates=[]` (inside the `if not pending_quit` branch) but, unlike the `ErrorMessage` handler just above (which sets `pending_quit=None`), it **keeps** any pre-existing `pending_quit`. Combined with the `MODEL_BUSY_TRANSITIONS` ordering, that's what routes a failed+already-quitting model to `ProcessPendingQuit` instead of `Done`.
2. When it clears `pending_updates` and the current state is `PendingMoreUpdates`, the forced `(Always, ProcessPendingUpdates)` then crashes (see the blocker comment).

Making this handler set `pending_quit = None` (or clearing it whenever `failed` is set) would mirror the `ErrorMessage` handler and resolve the `MODEL_BUSY_TRANSITIONS` divergence; the `PendingMoreUpdates` transition guards still need fixing for the crash.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -491,3 +467,69 @@ def determine_interdependency(self):
...
+    initial_state=Registration,
+    states={
+        Registration: MODEL_BUSY_TRANSITIONS,
+        NewTime: MODEL_BUSY_TRANSITIONS,
+        Updating: MODEL_BUSY_TRANSITIONS,
+        Idle: [
+            (HasPendingQuitAndIdle, ProcessPendingQuit),
+            (HasPendingUpdatesAndIdle, ProcessPendingUpdates),
+            (HasPendingNewTimeAndIdle, NewTime),
+        ],
```
onlyname0, line 527: Minor / consistency: `MODEL_BUSY_TRANSITIONS` includes `(ModelHasFailed, Done)` but the `Idle` list does not. Benign in the default config (a model only becomes `failed` while Idle via `handle_invalid_response`, which also sets `pending_quit`, so `HasPendingQuitAndIdle` catches it). But it's a composability footgun: a custom handler that sets `failed=True` in an Idle-family state without also setting a pending flag would leave the model parked in `Idle` (no matching condition → `next_state()` returns `None` → stays put) rather than reaching `Done`. Adding `(ModelHasFailed, Done)` here (and to `PendingMoreUpdates`) makes the declarative intent explicit and hardens custom configs.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
```diff
@@ -11,55 +12,76 @@
...
+            raise ValueError(
+                f"'initial_state' {self.initial_state.__name__} is not a member of 'states' "
+                "and 'strict' was set"
+            )
+        for _, state in itertools.chain.from_iterable(self.states.values()):
+            if state not in self.states:
+                raise ValueError(
+                    f"State {state.__name__} was mentioned in a transition but it is not a member "
+                    "of 'states' and 'strict' was set"
+                )
```
onlyname0, line 59: Composability footgun: `strict` validates state membership but not the central composition invariant — that a `requires_event=False` state implements `run()` and a `requires_event=True` state implements `handle_event()`. Since `run()` is no longer `@abstractmethod` (both default to `raise NotImplementedError`), a composer who sets `requires_event=False` (the default!) but only writes `handle_event` — or forgets `run` — builds a config that passes validation and then fails with a `NotImplementedError` deep inside `run_until_event_required`/`handle_event` at runtime. Consider validating the `requires_event` ↔ method-implemented contract here (and optionally that transition `Condition`s are `Condition` subclasses). Conditions themselves are currently never validated.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
```diff
@@ -68,40 +90,42 @@ def _run(self):
...
-        pass
 
-    @abstractmethod
     def run(self):
         raise NotImplementedError
 
-    def transitions(self) -> TransitionsT:
-        return []
+    def handle_event(self, event: t.Any):
+        raise NotImplementedError
```
onlyname0: Since composability is the point of the PR, a couple of contracts a custom-config author needs are currently undocumented and easy to get wrong — worth docstrings on `State`/`FSM`/`FSMConfig`:
- **Order matters**: `next_state()` returns the *first* matching condition, so transition lists are priority-ordered.
- **Resting states must be `requires_event=True`**: a run-only state that is reachable as a terminal/resting state (no firing transition, no raise) makes `run_until_event_required` spin forever with no diagnostic.
- **Catchable-exception contract**: only `FSMException` (and, at the orchestrator boundary, `InvalidCommand`) is handled; any other exception raised by a custom state crashes the orchestrator — which is exactly the mechanism behind the `PendingMoreUpdates` blocker.

packages/simulation-core/src/movici_simulation_core/exceptions.py
```diff
@@ -14,7 +14,7 @@ class OrchestratorException(SimulationException):
     pass
```
onlyname0, line 14: Nit: with `SimulationExit` removed, `OrchestratorException` now has no subclasses and no usages — it's dead. Either drop it, or (nicer) make `InvalidCommand` subclass `OrchestratorException` instead of `SimulationException` — that gives `InvalidCommand` the semantic precision suggested earlier and keeps this base class alive. (Just don't put it under `FSMException`, which would make `catch_fsmexceptions` swallow it.)

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -279,20 +303,24 @@ def _(self, msg: ResultMessage) -> None:
...
-    def notify_subscribers(self, command: t.Optional[Command] = None):
-        command = command or NoUpdateMessage()
+    def notify_subscribers(self, command: UpdateMessage | NoUpdateMessage):
+        """Notify subscribers that this ConnectedModel's model has returned."""
         for model in self.context.publishes_to:
+            # This method may technically raise InvalidCommand if any of the subscribed models
+            # does not accept an UpdateMessage or NoUpdateMessage, but the (default)
+            # ``ConnectedModel.fsm_config`` ensures that this cannot happen. In case of a custom
+            # ``fsm_config`` and this method raises ``InvalidCommand``, this is caught upstream by
+            # the orchestrator's ``Config``
```
onlyname0: Nit: "caught upstream by the orchestrator's `Config`" should read `Context` (it's `Context.recv_message` that catches `InvalidCommand`).

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: In `PendingMoreUpdates`, the model is idle. It is waiting for the orchestrator to give it a new command. A model is not supposed to send a message, although I guess an `ErrorMessage` could technically be allowed; although that requires changes on how we manage the zmq connection, so there's not an immediate possibility for a model to asynchronously send a message while Idle. The current model runner literally hangs in `zmq.Socket.recv()`. So no regression there. 

However, I do concede that we know about this possibility now, and we should at least document it while we have it fresh in our minds and that we don't loose track of this possibility in case we extend `Stream` to support more asynchronous communication. 

OTOH this is more about "a model crashes while it is being idle (or calculating)". I'd rather we solve it by detecting that a model died (using some kind of heart beating) and then terminate the simulation.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: When a model sends an invalid response as a result of being Busy, that means a programming error on the model's side. The model thinks it's doing the right thing, but isn't. In that case we need to tell it to terminate because it's misbehaving. So moving directly to Done is not the right thing. The only time we should not send that `QuitMessage` (ie directly move to `Done`) is if it crashed itself and it is no longer accepting commands.

This sounds like this refactor actually fixed a (previously unknown) bug, rather than create it. But i'll create a test and see how it behaves. 

Edit: test confirms that it was a bug that is fixed now :D

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: See above, the reason we mark the model as failed is because it misbehaved, not because it crashed itself. We actively need to tell it to stop by sending a QuitMessage

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: Like I said, there is currently no way that an idle model can spontaneously become `failed`. Good to keep in mind though for when we allow a more asynchronous approach for model runners.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
elfjes: Good point. I've added a check and some docstrings to communicate the requirements

packages/simulation-core/src/movici_simulation_core/services/orchestrator/fsm.py
elfjes: makes sense

packages/simulation-core/src/movici_simulation_core/exceptions.py
elfjes: Done, made InvalidCommand a subclass of OrchestratorException

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
elfjes: thanks :)

PR #138, opened 2026-06-10
Movici Data Core - Add service layer as a higher level api for interaction with data

elfjes:
Describe your changes
--------------------
Introduce a service layer in the movici data core backend. This service layer can perform additional tasks for interacting with data. For some interactions, this is not required and the service layer acts as a transparent proxy to the repository. Other interactions are more involved. Services are pluggable, so that specialized versions may be configured for an application

It's another big PR. I've tried separating the work in the PR in a number of commits that can each be read separately, to ease the review process. I hope that helps a bit

Not all service layers are complete yet, but the design is fixed enough to move to the next step (rest api)

some comments:
* introduces pydantic, but only as a data validation/serialization layer. I'd much rather use marshmallow for this, but that doesn't play as nice with fastapi, especially wrt generating openapi docs
* interaction with dataset- and update data is file based. This is done to facilitate multiprocessing parallelism (mostly relevant for the platform). A subprocess can prepare a file, that can subsequently be passed to a - future - http-server or framework and served to the client

Copyright license
-----------------


packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,382 @@
...
+            name=self.name,
+            type=ModelType(self.type) if isinstance(self.type, str) else self.type,
+            config=self.__pydantic_extra__ or {},
+        )
+
+
+class ScenarioModelOut(OutModel[ScenarioModel]):
+    model_config = ConfigDict(extra="allow")
+    name: str
+    type: t.Annotated[str, BeforeValidator(attrgetter("name"))]
```
onlyname0: **Blocker.** `BeforeValidator(attrgetter("name"))` (also line 260) makes `movici_data_core` unimportable on Python 3.11/3.12. With the lockfile's pydantic 2.12.5, pydantic inspects the validator signature, and `inspect.signature(attrgetter(...))` returns `(*args, **kwargs)` on 3.11/3.12, which pydantic rejects with `PydanticUserError` at class-definition time. The whole package then fails to import and the entire data-core suite dies at collection (reproduced: 3.10 OK via fallback, 3.11/3.12 fail, 3.13 OK). `requires-python` is `>=3.10,<3.14`; this slips through CI because the default PR job pins 3.13 and the full matrix is gated behind the `all-versions` label.

```suggestion
    type: t.Annotated[str, BeforeValidator(lambda v: v.name)]
```
(same fix at line 260). Worth running the `all-versions` matrix on this PR regardless.

packages/data-core/src/movici_data_core/bounding_box.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+        for attr in GEOMETRY_ATTRIBUTES:
+            if attr not in entity_data:
+                continue
+            data_array = _get_data(entity_data, attr)
+            min_x, max_x = DataType(float).get_min_max(data_array[:, 0])
+            min_y, max_y = DataType(float).get_min_max(data_array[:, 1])
+
+            bounding_box = calculate_new_bounding_box(
+                bounding_box, BoundingBox(min_x, min_y, min_y, max_y)
```
onlyname0: **Blocker — geometry corruption.** The third positional arg should be `max_x`, not `min_y`:

```suggestion
            bounding_box = calculate_new_bounding_box(
                bounding_box, BoundingBox(min_x, min_y, max_x, max_y)
            )
```
Every entity-based dataset or update containing `geometry.linestring_*`/`geometry.polygon*` silently persists a corrupt bounding box through both `DatasetService` and `UpdateService`. The unit test can't catch it (`test_bounding_box.py` uses values where the swap still passes, and a non-CSR layout). Please add a test with asymmetric x/y ranges and CSR data.

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,382 @@
...
+            display_name=self.display_name,
+            description=self.description,
+            epsg_code=self.epsg_code,
+            simulation_info=self.simulation_info.to_domain(),
+            datasets=[ds.to_domain() for ds in self.datasets],
+            models=[model.to_domain() for model in self.models],
+        )
+
+
+class SimulationInfoIn:
```
onlyname0: **Blocker.** `SimulationInfoIn` doesn't inherit from `BaseModel`, so any `ScenarioIn.model_validate(...)` raises at schema-build time. This is masked by the dead test at `test_schema.py:255` (see that comment). Once fixed, two more issues in this class surface: the expected `SimulationInfo(start_time=12, ...)` in the test references a field this class doesn't declare. Make it `class SimulationInfoIn(BaseModel):` and reconcile its fields with `SimulationInfo`.

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,382 @@
...
+            duration=self.duration,
+            reference=self.reference,
+            time_scale=self.time_scale,
+            mode=self.mode,
+        )
+
+
+class ScenarioDatasetIn(BaseModel):
+    name: str
+    type: DatasetType | None
```
onlyname0: Part of the `ScenarioIn` chain (blocker family): `type: DatasetType | None` rejects the plain-string form (`"type": "some_type"`) that scenario configs — and the PR's own fixture — use. `ScenarioModelIn` accepts `str | ModelType`; this should too (`type: DatasetType | str | None`, coercing the str in `to_domain`).

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def to_domain(self):
+        return domain_model.ScenarioModel(
+            name=self.name,
+            type=ModelType(self.type) if isinstance(self.type, str) else self.type,
+            config=self.__pydantic_extra__ or {},
+        )
+
+
+class ScenarioModelOut(OutModel[ScenarioModel]):
```
onlyname0, line 182: `ScenarioModelOut` only declares `name`/`type`; with `from_attributes=True` the model's `config` dict is never harvested, so serializing a `ScenarioModel` drops its entire config. Consider `model_config = ConfigDict(from_attributes=True, extra="allow")` plus an explicit mechanism to surface `config`, mirroring `ScenarioModelIn`.

packages/data-core/tests/test_schema.py
```diff
@@ -0,0 +1,272 @@
...
+                "reference": 9000.1,
+            },
+            "models": [
+                {"name": "model1", "type": "model_a", "dataset": "a_dataset"},
+                {"name": "model3", "type": "model_c"},
+            ],
+            "datasets": [{"name": "dataset_a", "type": "some_type"}],
+        }
+
+    @pytest.fixture
```
onlyname0: **This is a test, not a fixture** — the `@pytest.fixture` decorator means it's never collected, which is what hides the broken `ScenarioIn` chain (schema.py:136/153/178). Drop the decorator. Worth a `pytest --collect-only` sanity check after adding test classes.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -91,6 +173,52 @@ def for_workspace(self, workspace_id: UUID):
...
+        return self.scenario_service_cls(
+            self.repository, single_scenario_mode=self.single_scenario_mode
+        )
+
+    @property
+    def updates(self):
+        if self.serializer is None:
+            raise RuntimeError("SQLAlchemyBackend.serializer must be set")
+        if self.tmpfile_dir is None:
+            raise RuntimeError("SQLAlchemyBackend.tmpfile_dir must be set")
```
onlyname0, line 228: **Blocker (surfaced here).** The new `get_backend`/`_build_backend` path requires `default_scenario_id`/`default_workspace_id` and raises `InconsistentDatabase` otherwise. But `_upgrade_to_single_workspace_mode` / `_upgrade_to_multiple_workspaces_mode` (further down, untouched by this PR) clear the default scenario/workspace and flip the in-memory flags **without ever assigning `self.options.mode`** (the downgrade paths do). So after an upgrade + commit, the DB still records the old mode with its default cleared, and the next `get_backend()` raises here. `test_change_mode` passes only because it asserts in-memory flags and never reopens a backend. Please set `self.options.mode` in both upgrade paths and add a reopen-after-upgrade test.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -0,0 +1,206 @@
...
+        return outfile
+
+    async def get_entity_data(self, dataset_id: UUID):
+        return await self.repository.dataset_data.get_entity_data(dataset_id)
+
+    async def get_unstructured_data(self, dataset_id: UUID):
+        return await self.repository.dataset_data.get_unstructured_data(dataset_id)
+
+    async def stream_binary_data(self, dataset_id: UUID):
+        return self.repository.dataset_data.stream_binary_data(dataset_id)
```
onlyname0: **Major — missing `await`.** This returns the repository coroutine, not the async iterator, so `await service.stream_binary_data(id)` yields a coroutine and fails with `TypeError` when consumed. It's also the one service method with zero coverage. Compare the correctly awaited call in `get_dataset_as_file`.

```suggestion
        return await self.repository.dataset_data.stream_binary_data(dataset_id)
```

packages/simulation-core/src/movici_simulation_core/core/data_format.py
```diff
@@ -116,7 +129,7 @@ def dump_dict(self, dataset: dict):
 def load_from_json(
     data,
     schema: t.Optional[AttributeSchema] = None,
-    non_data_dict_keys=("general",),
+    non_data_dict_keys=NON_DATA_DICT_KEYS,
```
onlyname0, line 133: _(re: around line 135)_

**Major.** The dataclass conversion of `EntityInitDataFormat` dropped the old `None → AttributeSchema()` fallback from `__init__`, and `load_from_json` passes `schema` positionally — so a schema-less call (the documented default `schema=None`) sets `self.schema = None` and `load_attribute` later dies on `None.get_spec`. In-repo callers all pass a schema, but this is public API whose signature still advertises `Optional`. Restore the fallback (e.g. `schema or AttributeSchema()` here, or in `__post_init__`).

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -27,24 +43,39 @@ class SQLAlchemyServer:
...
     """
 
+    dbapi_url: str
+    tmpfile_dir: pathlib.Path
+    serializer: ExternalSerializationStrategy
+
     session_factory: async_sessionmaker[AsyncSession]
     engine: AsyncEngine
 
+    workspace_service_cls: type[WorkspaceService] = WorkspaceService
```
onlyname0, line 58: **Major — dead config.** These `SQLAlchemyServer.*_service_cls` attributes are never forwarded: `_build_backend` constructs `SQLAlchemyBackend(session, options, serializer=..., tmpfile_dir=...)` without them, so the advertised server-level pluggability silently does nothing — only the matching overrides on the `SQLAlchemyBackend` dataclass take effect. Either forward them in `_build_backend` or remove the server-level copies.

packages/data-core/src/movici_data_core/exceptions.py
```diff
@@ -145,7 +148,7 @@ def as_dict(self):
         return result
 
     def iter_messages(self):
-        prefix = self.path + "." if self.path else ""
+        prefix = f"{str(self.path)}." if self.path else ""
```
onlyname0, line 151: _(re: around line 156)_

**Major — int `path` half-finished.** `ensure_scenario_datasets` now raises `MoviciValidationError(..., path=idx)` with an `int`. Two problems: (1) `iter_messages` does `self.path + "."`-style prefixing — after `ScenarioRepository` wraps it via `from_errors(e, path="datasets")`, rendering crashes with `TypeError` on str/int concatenation; (2) at line 151 `if self.path` drops `path=0` entirely, so the first dataset's error loses its index. Normalize `path` to `str` in `__init__`.

packages/data-core/src/movici_data_core/database/repository/dataset.py
```diff
@@ -227,93 +208,145 @@ async def update_with_data(self, id: UUID, obj: Dataset, chunk_size=0):
...
+            existing_datasets.update(
+                (
+                    ds.name,
+                    ScenarioDataset(
+                        ds.name,
+                        dataset_type=existing_types[t.cast(DatasetType, ds.dataset_type).name],
+                        id=id,
+                    ),
+                )
+                for id, ds in zip(created_ids, to_create)
```
onlyname0: **Major — ordering assumption.** `insert(...).returning(db.Dataset.id)` with an executemany list correlates returned IDs to `to_create` purely by row order, which SQLAlchemy only guarantees with `sort_by_parameter_order=True`. Works on SQLite today; on other backends this can silently attach the wrong ID to the wrong dataset name (the base code mapped by name). Add the flag or map by name again.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -0,0 +1,206 @@
...
+class DatasetService:
+    def __init__(
+        self,
+        repository: SQLAlchemyRepository,
+        serializer: ExternalSerializationStrategy,
+        tmpfile_dir: pathlib.Path,
+    ):
+        self.repository = repository
+        self.serializer = serializer
+        if not tmpfile_dir.exists() and tmpfile_dir.is_dir():
```
onlyname0: Minor — this guard can never fire (precedence): `not exists() and is_dir()` is always False. Use `if not tmpfile_dir.is_dir():`. (`UpdateService` has no check at all.)

onlyname0 (commented):
Overall the architecture is sound and most of the PR is solid, but I don't think it should merge as-is: there are two outright blockers (a Python-version import crash and a geometry-corruption bug), two more bugs that leave the DB or public API broken, and a cluster of smaller defects — several invisible to CI because the one test covering the broken path is decorated `@pytest.fixture` and never runs, and the default PR job only runs Python 3.13.

**Verified:** data-core suite on the PR head (199 passed on 3.13; collection crashes on 3.11/3.12), simulation-core suite (1542 passed, 1 xfailed), and the pytest-cov→coverage migration works end-to-end (`make coverage`, 93% line coverage). The four headline bugs were reproduced empirically.

**Strengths worth calling out:** clean layering (domain dataclasses have no DB imports, repositories speak domain objects, services are thin async facades), a solid `ScenarioDataset`/`ScenarioModel`/`UpdateModel` refactor with field-by-field-correct db→domain mapping, mostly substantive integration tests, and a commit-per-concern structure that genuinely made this reviewable.

**Suggested merge gate:** fix the blockers + the dead test, the two majors in `stream_binary_data`/`load_from_json`, then the rest can be follow-ups (though the exceptions/returning/`*_service_cls` ones are cheap now). One process note: `pyproject.toml` adds `pydantic` with no version floor — given how sensitive this PR is to the exact pydantic version, a floor pin (e.g. `pydantic>=2.7`) would help.

packages/data-core/src/movici_data_core/services/update.py
```diff
@@ -0,0 +1,58 @@
...
+        self, update_id: UUID, filetype: FileType = FileType.JSON
+    ) -> pathlib.Path:
+        if filetype not in self.serializer.supported_file_types():
+            raise UnsupportedFileType(filetype)
+
+        result = await self.repository.updates.get_by_id(update_id, with_data=True)
+        if result is None:
+            raise ResourceDoesNotExist("update", id=update_id)
+
+        outfile = (
```
onlyname0: Minor — `.with_suffix()` mangles dataset names containing dots: `update-foo.bar-3-1-abcdef` loses everything after `foo.`, including the anti-collision random suffix → concurrent-export collisions. Append the extension with `+` instead. Same pattern in `DatasetService.get_dataset_as_file`.

packages/data-core/src/movici_data_core/database/repository/updates.py
```diff
@@ -115,7 +115,8 @@ async def create(self, obj: Update) -> UUID:
                     timestamp=obj.timestamp,
                     iteration=obj.iteration,
                     model_type_id=model_type_id,
-                    model_name=obj.model_name,
+                    model_name=obj.model.name,
+                    bounding_box=obj.bounding_box,
```
onlyname0: Minor — `bounding_box=obj.bounding_box` stores the raw `BoundingBox` NamedTuple. Geometry-less updates then persist as `[null,null,null,null]` instead of SQL `NULL`, defeating the new `isnot(None)` filter in `ScenarioRepository._get_bounding_box` (harmless today only because the union helper skips `None` components). Use `obj.bounding_box.as_tuple_or_none()` like the dataset path.

packages/data-core/src/movici_data_core/bounding_box.py
```diff
@@ -0,0 +1,55 @@
...
+        if GEOMETRY_Y in entity_data:
+            min_y, max_y = DataType(float).get_min_max(_get_data(entity_data, GEOMETRY_Y))
+            bounding_box = calculate_new_bounding_box(
+                bounding_box, BoundingBox(None, min_y, None, max_y)
+            )
+        for attr in GEOMETRY_ATTRIBUTES:
+            if attr not in entity_data:
+                continue
+            data_array = _get_data(entity_data, attr)
+            min_x, max_x = DataType(float).get_min_max(data_array[:, 0])
```
onlyname0, line 38: Minor — `data_array[:, 0]` raises `IndexError` for a CSR geometry attribute whose rows are all undefined (the data array is 1-D/empty in that case). Guard for empty/1-D before slicing.

packages/data-core/src/movici_data_core/services/update.py
```diff
@@ -0,0 +1,58 @@
...
+            self.tmpfile_dir / f"update-{result.dataset.name}-{result.timestamp}"
+            f"-{result.iteration}-{random_suffix()}"
+        ).with_suffix(filetype.default_extension)
+
+        UpdateWithDataOut.write_to_file(result, outfile, self.serializer, filetype)
+        return outfile
+
+    async def store_update_from_file(
+        self, path: pathlib.Path, filetype: FileType = FileType.JSON
```
onlyname0: Minor — `store_update_from_file` defaults `filetype=FileType.JSON` and always forwards it, so `UpdateIn.read_from_file`'s suffix-based detection can never trigger; a `.msgpack` upload without an explicit filetype is parsed as JSON. Default to `None` and let `read_from_file` infer.

packages/data-core/src/movici_data_core/services/scenario.py
```diff
@@ -0,0 +1,45 @@
...
+        self.single_scenario_mode = single_scenario_mode
+
+    async def list(self):
+        return await self.repository.scenarios.list()
+
+    async def get(self, name: str | None = None, id: UUID | None = None) -> Scenario | None:
+        if name is not None:
+            result = await self.repository.scenarios.get_by_name(name)
+        elif id is not None:
+            result = await self.repository.scenarios.for_id(id).get()
```
onlyname0, line 21: Minor — `get(id=...)` calls `scenarios.for_id(id)`, whose `_ensure_no_scenario_id()` raises whenever the repository is already scenario-bound (i.e. always in single-scenario mode), even for the matching id. Fall back to `scenarios.get()` when `repository.scenario_id == id`.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -0,0 +1,206 @@
...
+                #
+                # prevent pydantic from processing the data section
+                raw_data = DatasetWithDataOut.from_domain(
+                    dataclasses.replace(existing, data={})
+                ).model_dump(mode="json")
+                raw_data["data"] = data
+                outfile.write_bytes(dump_dict(raw_data, filetype=filetype))
+
+            case DatasetFormat.BINARY:
+                with open(outfile, "wb") as fp:
```
onlyname0: Minor — the BINARY branch opens the output file before streaming starts, leaving an orphaned empty temp file if streaming fails; it also ignores the `filetype` arg except for naming. Consider streaming to a temp path and renaming on success, and validating `filetype` up front.

packages/data-core/src/movici_data_core/database/repository/dataset.py
```diff
@@ -227,93 +208,145 @@ async def update_with_data(self, id: UUID, obj: Dataset, chunk_size=0):
...
         if to_create:
-            existing_types = {tp.name: tp for tp in await self.all_data.dataset_types.list()}
+            # to_create will only have entries after we have checked a dataset's dataset type and
+            # have retrieved ``existing_types`` so we can make the following assertion
+            assert existing_types is not None
             for scenario_dataset in to_create:
-                if scenario_dataset.type not in existing_types:
+                # A ScenarioDataset can only be added to to_create if it has a dataset_type
+                assert scenario_dataset.dataset_type is not None
+                if scenario_dataset.dataset_type.name not in existing_types:
```
onlyname0, line 323: Minor — when stub-creating a dataset whose type *name* already exists, a conflicting `format` on the supplied type is silently ignored rather than rejected (`is_equivalent` short-circuits on name when either format is None, but here the incoming format is set). Worth an explicit conflict check.

packages/data-core/src/movici_data_core/database/repository/dataset.py
```diff
@@ -227,93 +208,145 @@ async def update_with_data(self, id: UUID, obj: Dataset, chunk_size=0):
...
+                    }
+
+                if (
+                    scenario_dataset.dataset_type.name not in existing_types
+                    and self.options.STRICT_DATASET_TYPES
+                ):
+                    raise MoviciValidationError(
+                        {
+                            "type": [
+                                f"dataset type '{scenario_dataset.dataset_type}' does not exist"
```
onlyname0: Minor — this interpolates the whole `DatasetType` dataclass repr into the message: `f"dataset type '{scenario_dataset.dataset_type}' does not exist"`. Use `.name`.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -459,6 +482,9 @@ class Update(Base):
 
     timestamp: Mapped[int]
     iteration: Mapped[int]
+    bounding_box: Mapped[tuple[float, float, float, float] | None] = mapped_column(
```
onlyname0, line 485: Minor — the new `Update.bounding_box` column and `Scenario.epsg_code` becoming nullable change the DB schema without a schema-version bump or migration note. Flag for whatever migration story this project uses.

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,382 @@
...
+
+class WorkspaceIn(BaseModel):
+    name: str
+    display_name: str
+
+    def to_domain(self):
+        return domain_model.Workspace(name=self.name, display_name=self.display_name)
+
+
+class WorkspaceOut(WorkspaceIn, OutModel[domain_model.Workspace]):
```
onlyname0, line 77: Minor — `WorkspaceOut` requires `scenario_count`/`dataset_count` as `int`, but `WorkspaceService.get()` returns a `Workspace` with both `None` (only `list()`/`with_counts()` populate them). `WorkspaceOut.from_domain(get(...))` will raise a ValidationError — a footgun for the obvious REST handler. Either populate counts in `get`, or make these `int | None = None`.

packages/data-core/src/movici_data_core/serialization.py
```diff
@@ -0,0 +1,27 @@
...
+
+import msgpack
+import orjson
+
+from movici_data_core.exceptions import DeserializationError, UnsupportedFileType
+from movici_simulation_core.types import FileType
+
+
+def load_dict(data: bytes, filetype: FileType) -> dict:
+    try:
```
onlyname0, line 11: Minor — `load_dict` doesn't verify the result is a dict; a top-level JSON array parses fine here and escapes as `AttributeError` downstream instead of `DeserializationError`. Since this parses future client uploads, check `isinstance(result, dict)` inside the helper.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -0,0 +1,206 @@
...
+                "dataset",
+                id=dataset.id,
+                message=(f'Invalid mimetype. Expected "{dataset_type.mimetype}", got {mimetype}'),
+            )
+
+        return await self.repository.datasets.update(
+            dataset.id, obj=dataclasses.replace(dataset, data=path)
+        )
+
+    @staticmethod
```
onlyname0: Nit — `_ensure_supported_file_type` is dead code (never called).

packages/data-core/src/movici_data_core/database/repository/general.py
```diff
@@ -256,13 +259,15 @@ async def update(self, id: UUID, obj: ModelType):
         :param id: the UUID of the stored ``ModelType``
         :param obj: the ``ModelType`` object with the changes
         """
+        if obj.jsonschema is None:
+            raise InvalidAction("Cannot ModelType.jsonschema to None")
```
onlyname0: Nit — garbled message: `"Cannot ModelType.jsonschema to None"` (missing a verb).

packages/data-core/src/movici_data_core/domain_model.py
```diff
@@ -325,6 +371,8 @@ class Update:
         scenario must have a unique (timestamp, iteration) combination
     :param model_name: the name of the model in the scenario that produced the update
     :param model_type: the type of the model in the scenario that produced the update
+    :param bounding_box: The bounding_box of the update, in case it contains geospatial attributes.
```
onlyname0, line 380: _(re: around line 372)_

Nit — stale docstring after the field changes: this still documents `model_name`/`model_type` (now `model: UpdateModel`). A few sibling docstrings/messages are similarly stale after the renames (`ScenarioDataset.type`, `DatasetRepository.update`, `ensure_dataset_type`'s `UNKNOWN` reference, the `_store_scenario_details` typeless-model comment).

packages/data-core/tests/test_serialization.py
```diff
@@ -0,0 +1,31 @@
...
+
+from movici_data_core.exceptions import DeserializationError, UnsupportedFileType
+from movici_data_core.serialization import dump_dict, load_dict
+from movici_simulation_core.types import FileType
+
+
+def test_serialize_and_deserialize_json():
+    data = {"some": "data"}
+    assert load_dict(dump_dict(data, FileType.JSON), filetype=FileType.JSON)
+
```
onlyname0, line 11: Nit — these round-trips assert truthiness only (`assert load_dict(dump_dict(data))`), so a wrong-but-nonempty result still passes. Assert `== data`.

packages/data-core/tests/conftest.py
```diff
@@ -196,8 +198,7 @@ async def _create_update(timestamp, iteration, ids, array, scenario_id=None):
             dataset=ScenarioDataset(a_dataset.name, a_dataset.dataset_type.name),
             timestamp=timestamp,
             iteration=iteration,
-            model_name=a_scenario.models[0]["name"],
-            model_type=a_scenario.models[0]["type"],
+            model=UpdateModel(name=a_scenario.models[0].name, type=a_scenario.models[0].type),
```
onlyname0, line 201: _(re: around line 198)_

Nit — passes a `str` where `ScenarioDataset.dataset_type` now expects a `DatasetType`. Works by accident, but undermines the typed refactor this PR is making.

packages/data-core/src/movici_data_core/schema.py
elfjes: I had not realized that, thanks for the catch :D

packages/data-core/src/movici_data_core/bounding_box.py
elfjes: thanks :)

packages/data-core/src/movici_data_core/schema.py
elfjes: good catch :)

packages/data-core/src/movici_data_core/schema.py
elfjes: yep, agree

packages/data-core/src/movici_data_core/schema.py
elfjes: yeah, some shenanigans with BeforeValidator required...

ugh I'm starting to regret using pydantic. I'd much rather use Marshmallow, but the pydantic integration in fastapi (especially wrt openapi schema generation) is nice. Pydantic tries to be too many things, validation/serialization, domain model, even an orm model (SQLModel). If only there was a nice way to use Marshmallow with fastapi that supports openapi spec generation...

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: thanks

packages/data-core/src/movici_data_core/domain_model.py
elfjes: thanks :+1:

packages/data-core/tests/test_serialization.py
elfjes: sloppy :see_no_evil:

packages/data-core/tests/test_schema.py
elfjes: wow

packages/data-core/src/movici_data_core/database/backend.py
elfjes: nice catch, thanks!

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: Thnx, fixed and added some tests.

packages/simulation-core/src/movici_simulation_core/core/data_format.py
elfjes: fixed by making the `schema` param from `load_from_json` required instead of optional

packages/data-core/src/movici_data_core/database/backend.py
elfjes: fixed by forwarding the classes

packages/data-core/src/movici_data_core/exceptions.py
elfjes: thanks

packages/data-core/src/movici_data_core/database/repository/dataset.py
elfjes: thanks, I did not know that :D. decided against using the flag, since it may do black magic in the background but made sure to let the insert return the dataset name as well, so we can map by name

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: thanks, fixed and also added the guard to updateservice

packages/data-core/src/movici_data_core/services/update.py
elfjes: dataset names are not allowed to have dots, only lowercase and underscores are allowed. Validation is still on my todo. However, it makes sense to be defensive and fix this now, in case we decide to relax this requirement in the future

packages/data-core/src/movici_data_core/database/repository/updates.py
elfjes: thanks

packages/data-core/src/movici_data_core/bounding_box.py
elfjes: thanks for the catch. Fixed by checking for an empty array. Not checking the shape though. At some point we should make sure that certain attribute types are always defined with a specific data type in the AttributeSchema / database (such as 'id' and the geometry types). Not today though, for now we assume that the geometry types are defined as expected

packages/data-core/src/movici_data_core/services/scenario.py
elfjes: thanks, fixed by allowing "changing" to the current id

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: temporary file management is still a todo. Things to consider:

* clean up files after a failed call (your suggestion here)
* clean up files after the `SQLAlchemyServer` closes (for shorter running servers)
* clean up files periodically (for longer running servers) <- this is probably out of scope for movici-data-core
* manually clean up files (for example after a fastapi request)

solved the first one by introducing a function that wraps around `tempfile.NamedTemporaryFile`. Others are out of scope for this PR

packages/data-core/src/movici_data_core/database/repository/dataset.py
elfjes: Not sure I follow. An existing dataset type will always have a format set, and if the dataset stub type does not have a format set, a default format (entity_based) is picked. So we always compare two dataset types with a format and never short-circuit to checking only by name. Or am I missing something?

packages/data-core/src/movici_data_core/database/model.py
elfjes: Data core is still in pre-release stage. No migrations required yet

packages/data-core/src/movici_data_core/schema.py
elfjes: This is as intended. We need a cheap call to retrieve a workspace, at the very least to check whether it exists during REST operations and authorization. Since getting counts is an additional/expensive operation, it should be possibly to opt-in/opt-out from getting the counts. Which is what the `with_counts` method was for. I've refactored it to become a `get_with_counts` method, to be a bit more symmetrical with `get`.

The WorkspaceOut BaseModel exists to make sure that the REST operations does the right thing and returns the right data, ie. a workspace with counts

packages/data-core/src/movici_data_core/serialization.py
elfjes: good point

elfjes:
Thanks for the thorough review, very helpful :). omg there were some really stupid mistakes there :see_no_evil:

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,414 @@
...
+class UpdateWithDataOut(ShortUpdateOut):
+    data: dict | None
+
+    @classmethod
+    def write_to_file(
+        cls,
+        update: Update,
+        path: pathlib.Path | t.BinaryIO,
+        serializer: ExternalSerializationStrategy,
+        filetype: FileType,
```
onlyname0, line 302: Re-review of `4c1e474` (new): making `filetype` required is good, but removing the internal `if filetype not in serializer.supported_file_types(): raise UnsupportedFileType(filetype)` guard means a direct caller of this public classmethod now gets a generic `TypeError` from `serializer.dumps` instead of the domain-specific `UnsupportedFileType`. The only in-repo caller (`UpdateService.get_update_as_file`) validates up front, so no current impact — but the docstring below still promises the filetype "must be a FileType that is supported by the serializer", and the sibling methods (`DatasetWithDataIn.read_*`, `UpdateIn.read_from_file`) keep their guards. Either re-add a one-line `serializer.supported_file_type_or_raise(filetype)` guard, or soften the docstring to say validation is the caller's job.

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,414 @@
...
+class ScenarioModelOut(OutModel[ScenarioModel]):
+    model_config = ConfigDict(extra="allow")
+    name: str
+    type: t.Annotated[
+        str | ModelType, BeforeValidator(lambda v: v.name if isinstance(v, ModelType) else v)
+    ]
+
+    @classmethod
+    def from_domain(cls, obj: ScenarioModel):
+        return ScenarioModelOut(name=obj.name, type=obj.type, **obj.config)
```
onlyname0, line 191: Re-review of `4c1e474` (nit): `ScenarioModelOut(name=obj.name, type=obj.type, **obj.config)` raises `TypeError: got multiple values for keyword argument 'name'` if `obj.config` ever contains a `name` or `type` key. Unreachable through the current flow (input routing keeps them out of `config`, and `_stripped_config` pops them before persistence), so this is hardening for a public classmethod rather than an active bug. `cls.model_validate({**obj.config, 'name': obj.name, 'type': obj.type})` avoids the collision and makes the explicit fields win.

packages/data-core/src/movici_data_core/schema.py
```diff
@@ -0,0 +1,414 @@
...
+        update: Update,
+        path: pathlib.Path | t.BinaryIO,
+        serializer: ExternalSerializationStrategy,
+        filetype: FileType,
+    ):
+        """
+        Serialize and write an Update to a file.
+
+        :param update: the Update to write
+        :param file: Either a ``pathlib.Path`` or a file-like object. If given a file-like object
```
onlyname0, line 308: Re-review of `4c1e474` (nit): the docstring documents `:param file:`, but the parameter is `path` (line 300). Rename the docstring entry to `:param path:` (or rename the parameter to `file` to match the prose).

onlyname0 (commented):
Re-review of `4c1e474` — follow-up items. (The blocker/major fixes from the first round all check out; 224 tests pass on Python 3.10/3.12/3.13.) Three small new items in the schema code below; I've also replied in-thread on the bounding-box and dataset-type-format threads.

packages/data-core/src/movici_data_core/bounding_box.py
onlyname0: Following up: the shape check does matter, because the crashing 1-D array comes from the serializer itself, not from a misconfigured `AttributeSchema`. When a geometry attribute is present but every row is undefined, `parse_csr_list` produces a 1-D `(N,)` array (not `(N, k)`), so `len() > 0` passes the new empty-guard and `data_array[:, 0]` still raises `IndexError`. Reproduced through `EntityInitDataFormat` with the correct CSR schema:

```
geometry.polygon_3d = [None, None]  ->  data.shape (2,)  ->  data_array[:, 0] IndexError
```
Defined geometry (shape `(N, 3)`) is fine, so it's an edge case and totally reasonable to defer — just flagging that it's a real crash on valid serializer output reachable via `store_update_from_file` / `update_from_file`, not only a "geometry not defined as expected" case. `if not len(data_array) or data_array.ndim < 2: continue` closes it if you want.

packages/data-core/src/movici_data_core/database/repository/dataset.py
onlyname0: Good question — you're right that the common path is safe; the case I mean is an *explicitly* conflicting format in the input, not a missing one. `ScenarioDatasetIn.type` accepts a full `DatasetType` (as a dict) including a `format`, so a caller can pass e.g. `{"name": "bar", "format": "binary"}`. If a type named `bar` already exists as `entity_based`, this `to_create` branch finds `bar` in `existing_types`, reuses its id (lines 327-329) and the requested `binary` is silently dropped — there's no `is_equivalent` check here, unlike the already-exists *dataset* branch above (line 296). For the string-type / format-`None` path you described there's no conflict, so if explicitly-conflicting formats just aren't a supported input, feel free to ignore — it's narrow.

elfjes: I see, yes. I'm adding a check for this to raise an error in case of an incorrect dataset format

packages/data-core/src/movici_data_core/bounding_box.py
elfjes: I see. This is actually a bug in simulation-core that should not be allowed to create an array with the incorrect shape (which it currently does in certain cases). I'll write up a fix for that in a different PR.

I've created #142 (and merged it in specifically in this PR to fix the tests) to address this issue

packages/data-core/src/movici_data_core/schema.py
elfjes: sure, makes sense :)

packages/data-core/src/movici_data_core/schema.py
elfjes: If config contains a `"name"` or "`type"` field, something has gone wrong. Those are not supposed to be allowed in there. I'm adding a check in `ScenarioModel` to catch this at the root

packages/data-core/src/movici_data_core/schema.py
elfjes: thnx :)

elfjes:
Current state of the PR depends on #142

PR #137, opened 2026-06-08
Add urban drainage (SWMM/pyswmm) simulation model package

onlyname0:
## Summary

Adds `movici-urban-drainage-model`, a new workspace package that integrates the EPA SWMM engine via [pyswmm](https://www.pyswmm.org/), mirroring the structure of the existing drinking-water (WNTR) package.

It models urban drainage (storm water and sewer) networks: dynamic-wave hydraulic routing plus rainfall-runoff hydrology over subcatchments.

## What's included

- **Entity groups** for junctions, outfalls, storage units, conduits, pumps, orifices, weirs, outlets, subcatchments and rain gages (`attributes.py`, `dataset.py`).
- **`Model`** (`TrackedModel`, plugin name `urban_drainage`) that keeps a single live pyswmm `Simulation` open and advances it forward to each Movici moment, applying control inputs (rainfall via rain gages, node inflow, regulator target settings) *before* stepping — since SWMM marches forward and cannot rewind.
- **Runtime `.inp` synthesis** from the Movici dataset (`inp_writer.py` + the wrapper), and a per-step read-back of depth / head / flooding / flow / runoff into `PUBLISH` attributes (`simulation_wrapper.py`).
- **`swmm` dataset-creator source** that parses an existing `.inp` into Movici entities and geometry, dependency-free (`swmm_source.py`).
- Workspace wiring (dependency, `tool.uv.sources`, ruff isort / per-file-ignores).

## Design note

Unlike WNTR (which runs a full simulation per call), pyswmm drives a live, forward-only iterator. The wrapper therefore keeps one persistent `Simulation` across all `update(moment)` calls and reads results per-object by id (no results DataFrame).

## Tests

`make unittest` passes (50 tests). Coverage spans the model lifecycle and scheduling, the simulation wrapper (`.inp` synthesis, stepping, control inputs, every node/link type including tabular curves, multi–rain-gage routing, end-of-horizon handling), and the `.inp` source parser with a dataset-creator integration test.

## Scope notes

- **Dividers** are not modelled (SWMM treats them as ordinary junctions under dynamic-wave routing).
- Link **velocity** is not published (pyswmm exposes no usable per-link velocity / flow-area at runtime); the Froude number is published instead.
- Outfalls support FREE / NORMAL / FIXED (TIDAL / TIMESERIES require external boundary data and raise a clear error).

PR #139, opened 2026-06-11
Add enum value/string helpers to AttributeOptions

onlyname0:
## Summary

Adds two small helpers for converting between an enum attribute's keyword string and its integer index, on `AttributeOptions` and proxied on `Attribute`:

- `get_enum_value(string, case_sensitive=True) -> int`
- `get_enum_string(value) -> str`

Both raise a clear `ValueError` when no enumeration is configured, the string is not a member, or the index is out of range. `get_enum_value` accepts a `case_sensitive` flag for case-insensitive lookups.

## Motivation

These came out of the review of the urban-drainage model PR (#137): models that map an integer enum attribute to/from a SWMM/engine keyword currently either hand-roll `attribute.options.enum_values[i]` or build a throwaway class via `get_enumeration()`. A pair of explicit, validated methods is cleaner and reusable (drinking-water and urban-drainage both do this kind of conversion). `get_enumeration()` is left in place (it's only used in tests); it can be deprecated/removed in a follow-up if desired.

## Tests

Added cases to `TestEnum` in `tests/data_tracker/test_attribute.py` covering value↔string conversion, the `case_sensitive` flag, unknown-member and out-of-range (incl. negative) errors, and the no-enumeration error. `test_attribute.py` passes (86 tests).

### DIFF
--- a/packages/simulation-core/src/movici_simulation_core/core/attribute.py
+++ b/packages/simulation-core/src/movici_simulation_core/core/attribute.py
@@ -116,6 +116,41 @@ class AttributeOptions(t.Generic[T]):
             )
         return self._enum
 
+    def get_enum_value(self, string: str, case_sensitive: bool = True) -> int:
+        """Return the integer index of the enum member *string*.
+
+        :param string: the enum keyword to look up
+        :param case_sensitive: when ``False``, match case-insensitively
+        :raises ValueError: if no enumeration is configured or *string* is not a member
+        """
+        if self.enum_values is None:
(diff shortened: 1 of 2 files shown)


elfjes (approved):
nice, approved

alternatively, as an optimization, you could build a case_sensitive and case_insensitive lookup dict when the enum_values are set (convert enum_values to `@property`) so that you don't need to iterate over it multiple times. But probably very minor gains because it contains only a few items

Issue #127, opened 2026-04-30
Proposal: transparent attribute mapping for conflict resolution, predictor-correctors and iterative relaxation


elfjes:
In an ideal simulation, there is only one publisher per (dataset, entity group, attribute-array)-combination (from now on just: attribute) and we have an directed, acyclic dependency graph (DAG) so that no cycles exist. Most Movici simulations currently have this behaviour, but not all. Also, future use cases may involve different scenarios where we have multiple models publishing to the same attribute (leading to conflicts) or we need to modify a model's input or output using predictor-corrector algorithms or PID-controllers in order to speed up or effect convergence. Our converation with @averbraeck indicated that we need to prepare for these use cases

## Current approach
Our current approach in case of a conflict or a cyclic depencency is to first modify the model's wrapper so that the attribute at which the model subscribes to data or publishes their result  can be configured. For example the `traffic_demand_calculation` model can be configured to subscribe to `transport.average_time.star` instead of `transport.average_time` for calculations involving an induced demand effect. We then introduce a convergence model (implemented using a `udf` model) to (gradually and iteratively) update the `transport.average_time.star` based on the `transport.average_time` result.

A conflict may occur when we have two models publishing to the same attribute. Keeping in line with the above example, two instances of the `traffic_demand_calculation` may want to publish to the `transport.cargo_demand` attribute, one for international cargo demand and one for domestic demand. We currently have to configure the models to publish to different attributes, ie `transport.international_cargo_demand` and `transport.domestic_cargo_demand` and combine the two in a separate model, usually another `udf` model.

These methods all involve explicit mapping of the desired attribute to an intermediate, temporary, attribute. This is a manual process which depends on model(wrapper) support for changing the input or output attribute based on configuration. By formalizing and automating this process we can make this more transparent to the models and support more use cases in a structured but flexible way.

## Proposal

When dealing with one of these situations, a scenario developer includes one or more models that perform a certain numerical role in the simulation, let's call them "solver helpers" for now. These include functionalities like

* relaxation (update an attribute gradually based on a proposed new value)
* predictor-corrector (update an attribute by making a prediction on what the next value will be based on historical values (derivative))
* combiner (combine the result of two or more conflicting results in a configurable way (min, max, avg, sum, ...)
* track mulitple conflicting results, accept one of them as the canonical value, and track the divergence of the other values from the canonical value

[...]

PR #140, opened 2026-06-17
Transparent attribute mapping (REMAP) for conflict resolution and solver helpers (#127)

onlyname0:
Describe your changes
--------------------

Implements the transparent attribute-mapping (REMAP) design from #127: a structured way to
let multiple models publish the same attribute (conflict resolution) or break cyclic
dependencies (relaxation / predictor-corrector), without any model needing to know its
attributes were rerouted.

### Mechanism

* Models register with an optional `priority` on their `READY` message (`Priority.REGULAR=10`,
  `Priority.SOLVER_HELPER=20`; any non-negative int is allowed). The orchestrator grants
  ownership of each attribute to the unique highest-priority publisher; a tie at the top is a
  conflict and fails the run with an actionable error.
* Non-owner publishers are transparently remapped to an internal variant name
  (`<attr>:<model>:i`). The owner publishes the canonical name. Chains (`10 → 20 → 30`) route
  level-by-level.
* A new `REMAP` command is sent once, after all `READY`s and before the first `NEW_TIME`. The
  `ModelAdapter` installs rename middleware so publication/subscription stays transparent;
  solver helpers that consume the per-publisher variants implement an optional `remap()` hook.
* Internal `:i` variants are filtered out of wildcard (`*`) subscriptions, so data collectors
  don't record them; explicit subscription still returns them.

### What's here (one commit per slice)

1. `READY` priority field + `RemapMessage` + `Priority` enum
2. Pure ownership planner (`compute_remap_plan`) + `:i` naming
3. Wildcard `:i` filtering + shared mask-remap helpers
4. Connector REMAP middleware + `Model.remap()` adapter hook
5. Orchestrator FSM integration (`AwaitingRemap`/`Remapping` states, conflict → clean shutdown)
6. A minimal `combiner` solver helper (sum/mean/min/max) + end-to-end test

### Testing

* Full suite green (`NUMBA_DISABLE_JIT=1 ... pytest`): 1627 passed, 1 xfailed; ruff clean.
* Unit coverage for the planner (ties, multi-level chains, back-propagation), the connector
  rename middleware, the wildcard filter, and the per-model FSM.
* End-to-end (in-process runner): two models conflict on one attribute → a `combiner` takes
  ownership → the canonical mean is produced and **no `:i` variants leak** to a wildcard data
  collector. A second e2e proves a solver-helper `remap()` failure terminates the run.

### Out of scope (deferred per the #127 discussion)

Divergence monitoring / `DivergenceTracker`; per-attribute priority; runtime ownership transfer
(REMAP is one-shot, ownership fixed for the run); `include_private_attributes` GET flag;
cumulative-vs-instantaneous attribute semantics (belongs to the metadata work); CSR aggregation
in solver helpers (the combiner rejects CSR explicitly).

### Notes for review

* No distributed-path (multi-process ZMQ) e2e test yet — everything verified end-to-end runs
  in-process. The distributed REMAP handshake is wired and unit-covered, but a `distributed=True`
  e2e is a worthwhile follow-up.

Copyright license
-----------------


packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def initialize(self, state: TrackedState):
+        return
+
+    def update(self, state: TrackedState, **_):
+        if self._emitted:
+            return None
+        self._attr.array[:] = self._value
+        self._emitted = True
+        return None
```
elfjes, line 39: the _emitted logic is a bit distracting here. TrackedModelAdapter already takes care that the updated is only produced once, so no need to have an _emitted check here

packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
```diff
@@ -0,0 +1,201 @@
...
+        if "cargo_demand" not in eg:
+            continue
+        values = eg["cargo_demand"]
+        # The DataCollector's snapshot may serialise as {"data": [...]} or as a bare list.
+        if isinstance(values, dict):
+            values = values.get("data", [])
+        if values and all(abs(v - 20.0) < 1e-9 for v in values):
+            found_combined = True
+            break
+    assert found_combined, f"expected mean(10, 30) == 20.0 in canonical output, saw: {saved}"
```
elfjes: I think we know exactly what files we are looking for and what their content should be. In fact, I think there should only be a single update file: the one coming from the combiner. The other updates from the emitters should be completely dropped because they are empty.

It probably most obvious to just try and read that file and check its contents

packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # Suppress an "unused" warning on the logger import — kept available in case the test
+    # is extended to assert on orchestrator log lines.
+    logging.getLogger("movici").debug("REMAP e2e test completed")
+
```
elfjes: Seems redundant. If we want to add such a test, we can always reimport, no?

packages/simulation-core/src/movici_simulation_core/model_connector/connector.py
```diff
@@ -122,6 +163,77 @@ def close(self, message: QuitMessage):
...
+            if not eg_renames:
+                out[ds][eg] = attrs
+                continue
+            if not isinstance(attrs, dict):
+                raise ValueError(
+                    f"REMAP {side} rename: malformed update at '{ds}/{eg}'; "
+                    f"expected dict of attributes, got {type(attrs).__name__}"
+                )
+            out[ds][eg] = {eg_renames.get(name, name): value for name, value in attrs.items()}
+    return out
```
elfjes: I think you can reduce this to:

```suggestion
def _apply_renames(data: dict, mapping: dict):
    def _helper(data: dict, mapping: dict, level: int):
        if level == 3:
            return {mapping.get(k, k): v for k, v in data.items()}
        return {
            k: (_helper(v, mapping[k], level=level + 1) if k in mapping else v)
            for k, v in data.items()
        }

    try:
        return _helper(data, mapping, level=1)
    except (ValueError, TypeError):
        raise ValueError("Malformed data or mapping") from None
```

True, you'd lack a little in error messages, but this is an internal method anyway, and both the data and the mapping are (should be) programmatically generated, so already tested to be of correct shape

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+    output: t.Optional[AttributeObject] = None
+    inputs: t.List[AttributeObject]
+
+    def __init__(self, model_config: dict):
+        super().__init__(model_config)
+        method = self.config.get("method", "sum")
+        if method not in _METHODS:
+            raise ValueError(
+                f"combiner: unsupported method '{method}'. Choose one of: {sorted(_METHODS)}"
+            )
```
elfjes, line 82: This benefits from a jsonschema

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+    def setup(
+        self,
+        state: TrackedState,
+        schema: AttributeSchema,
+        **_,
+    ):
+        self.state = state
+        self.schema = schema
+        attribute = self.config["attribute"]
+        spec = schema.get_spec(attribute["name"], default_data_type=DataType(float))
```
elfjes, line 95: I think we can test for `csr=False` here to limit ourselves to UniformAttributes

Also we can do additional validations. Such as what about higher dimensional shapes?

Do we support all methods for all data types?
allowed methods for bool? Perhaps only min and max? although sum would be equivalent to `max`. And with proper rounding, `mean` could result in a democratic consensus

What about string types?

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+        """Register the internal-variant attribute fields the orchestrator instructs us to
+        subscribe to. We return ``False`` so the adapter still installs the rest of the
+        REMAP plumbing (pub-side, sub-side if it were one-to-one), but the connector will
+        skip the sub-rename middleware on its own because the sub remap is many-to-one."""
+        sub = payload.get("sub") or {}
+        for ds, entity_groups in sub.items():
+            for eg, mapping in entity_groups.items():
+                for variant in mapping.keys():
+                    spec = self.schema.get_spec(variant, default_data_type=DataType(float))
+                    attr = self.state.register_attribute(ds, eg, spec, flags=OPT)
```
elfjes: If you want to be sure that update is only called once the data is available, register as `SUB` here instead of `OPT`

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+    def remap(self, payload: dict) -> t.Optional[bool]:
+        """Register the internal-variant attribute fields the orchestrator instructs us to
+        subscribe to. We return ``False`` so the adapter still installs the rest of the
+        REMAP plumbing (pub-side, sub-side if it were one-to-one), but the connector will
+        skip the sub-rename middleware on its own because the sub remap is many-to-one."""
+        sub = payload.get("sub") or {}
+        for ds, entity_groups in sub.items():
+            for eg, mapping in entity_groups.items():
+                for variant in mapping.keys():
+                    spec = self.schema.get_spec(variant, default_data_type=DataType(float))
```
elfjes: I think we can reuse+rename the original spec since they will be equal. Someone who deliberately registers an internal attribute in the schema with a different spec should be shot on sight

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+            return None
+        stacked = np.stack([attr.array for attr in self.inputs])
+        combined = _METHODS[self._method](stacked)
+        if not isinstance(self.output, UniformAttribute):
+            return None
+        if not self.output.has_data():
+            # The output's backing array tracks the entity-group length, which is set when
+            # any update touches the group. By the time we have data on every input we are
+            # guaranteed the entity count, so initialise the output to match.
+            self.output.initialize(len(combined))
```
elfjes: `self.output.has_data()` should always return `True` at this point. If not, that might be something we have to address in `TrackedState`

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
```diff
@@ -0,0 +1,137 @@
...
+        elif np.array_equal(self.output.array, combined):
+            return None
```
elfjes: We can leave testing for changes to the `TrackedState`

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
```diff
@@ -350,6 +372,39 @@ def on_enter(self):
...
 
+    def _transitions(self):
+        # After receiving a RegistrationMessage, transition to AwaitingRemap (instead of
+        # the regular Idle) so the orchestrator can optionally send a REMAP command before
+        # the first NewTime. See issue #127.
+        return [
+            (lambda c: c.failed, Done),
+            (lambda c: (not c.busy) and c.quit, ProcessPendingQuit),
+            (lambda c: not c.busy, AwaitingRemap),
+        ]
```
elfjes: let's get that composable fsm PR merged first so we don't have to do this twice

elfjes (commented):
In general this looks good. Aside from the comments below, I've taken the liberty to directly implement some changes that I think are beneficial to the code:

* moved Priority to `types.py
* cast Priority to int in RegistrationMessage so that we can use Priority levels throughout the app (without specific casting and still be sure to have a plain int to serialize
* Model.remap returns explicit remap instructions (automatically install pub, sub or dual remap. This `AutoRemap` is a renamed variant on `RemapDecision`
* Introduced RemapMiddleware class that contains remap logic, and simplify the rename logic. compared to your implementation the error messages are a bit less verbose, but since the update data structure and the remap contents are (/should be) generated by the framework, we don't expect a model developer to be able to trigger that error, so a concise error suffices.
* deduplicated logic in `apply_remap_to_pub_mask` and `remap_to_sub_mask` and combined into a single `apply_remap_to_data_mask` function
* data mask wildcards can occur at different levels (at root, inside a dataset or inside an entity group). the `apply_remap_to_data_mask` function now reflects that
* added some additional tests

packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
onlyname0: Right — removed. The emitter now just writes its value every update and the change tracking makes sure it's only published once.

onlyname0:
Thanks for the direct improvements — `AutoRemap`, the `RemapMiddleware` extraction and the unified `apply_remap_to_data_mask` are all clearly better than what they replaced, and I've kept them as-is. All remaining comments are now addressed (details in the individual threads).

A few things I fixed on top while going through the branch:
- Two leftovers of the refactor: the empty `core/priority.py` and `tests/core/test_attribute_remapping.py`, which still imported the removed `core.attribute_remapping` module and broke pytest collection.
- The many-to-one `RemapError` message advised returning `AutoRemap(sub=True)` — the very thing that triggers the error; it now advises `sub=False`. Also rewrote the `remap()` docstrings that still described the old `None`/`True`/`False` protocol.
- New planner guard: a model that must *receive* `:i` variants but subscribes with a wildcard can never get them (wildcard delivery filters internal variants), so `compute_remap_plan` now raises `WildcardSubscriberError` instead of producing a plan that silently doesn't work.
- The e2e tests are now parametrized over `distributed=[False, True]`, so the ZMQ REMAP handshake is covered end-to-end and the caveat in the PR description is gone.

packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
onlyname0: You were right on both counts: the emitters' updates are dropped entirely (after the wildcard filter strips their `:i` variants nothing is left), so exactly one file is written. The test now asserts the directory contains only `t0_0_the_dataset.json` and compares its full contents — which also proves no `:i` variant leaked.

packages/simulation-core/tests/simulation/integration/test_remap_e2e.py
onlyname0: Removed, along with the `logging` import.

packages/simulation-core/src/movici_simulation_core/model_connector/connector.py
onlyname0: Kept your `RemapMiddleware` version; only fixed the level comment ("level 3" → "level 2").

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: Added `json_schemas/models/combiner.json` (draft-07, `movici.type` annotations, `method` as an enum), wired via `__model_config_schema__`. The runtime `ValueError` stays for `validate_config=False` construction.

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: The output spec is now validated at setup: CSR and string attributes are rejected, and bool is restricted to `min`/`max` (as you say, `sum` ≡ `max`; I left the rounded-`mean` consensus idea out for now). Since the variants reuse the output spec (see the other thread), validating the output covers the inputs too. Higher-dimensional shapes work as-is — the reductions run over `axis=0` of the `(n_inputs, n_entities, *unit_shape)` stack. One addition: the `mean` of integer inputs is rounded via `np.rint` instead of being truncated by the cast.

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: Done, and it turned out to solve more than the ordering: since `is_initialized()` requires every entity to be defined, the adapter now defers `update()` until every input variant is *fully* defined — so the combiner can never aggregate an undefined sentinel. Documented in the module docstring and covered by `test_partially_undefined_input_defers_output`.

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: Done — `dataclasses.replace(output_spec, name=variant)`, no more float inference (and no more shooting on sight required). Identity entries (`{canonical: canonical}` from back-propagation) are now skipped as well, so the combiner can't register its own output as an input.

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: Confirmed — by the time `update()` runs the entity group has received data and `TrackedState` has initialized the output array, so the guard was dead code. Removed; nothing to address in `TrackedState`.

packages/simulation-core/src/movici_simulation_core/models/combiner/combiner.py
onlyname0: Removed the `np.array_equal` check — `TrackedArray` compares against its snapshot on publish, so writing equal values produces no update.

packages/simulation-core/src/movici_simulation_core/services/orchestrator/context.py
onlyname0: #135 is in main now, so I merged main into this branch and ported the REMAP integration to the composable style: `ConnectedModel` gets a `pending_remap` slot with a `HasPendingRemapAndIdle` condition (mirroring `pending_new_time`), `AwaitingRemap`/`Remapping` are wired into `MODEL_FSM_CONFIG` (Registration parks models in `AwaitingRemap`; REMAP stays one-shot — after the ack a model settles in plain `Idle`), and `ComputeAndSendRemap`/`WaitForRemapAcks` are declared in `service.py`'s `FSM_CONFIG`. No `transitions()` methods left on states.

elfjes:
> Thanks for the direct improvements — `AutoRemap`, the `RemapMiddleware` extraction and the unified `apply_remap_to_data_mask` are all clearly better than what they replaced, and I've kept them as-is. All remaining comments are now addressed (details in the individual threads).
> 
> A few things I fixed on top while going through the branch:
> 
>     * Two leftovers of the refactor: the empty `core/priority.py` and `tests/core/test_attribute_remapping.py`, which still imported the removed `core.attribute_remapping` module and broke pytest collection.
> 
>     * The many-to-one `RemapError` message advised returning `AutoRemap(sub=True)` — the very thing that triggers the error; it now advises `sub=False`. Also rewrote the `remap()` docstrings that still described the old `None`/`True`/`False` protocol.
> 
>     * New planner guard: a model that must _receive_ `:i` variants but subscribes with a wildcard can never get them (wildcard delivery filters internal variants), so `compute_remap_plan` now raises `WildcardSubscriberError` instead of producing a plan that silently doesn't work.
> 
>     * The e2e tests are now parametrized over `distributed=[False, True]`, so the ZMQ REMAP handshake is covered end-to-end and the caveat in the PR description is gone.

nice! I see there's an issue with a docstring, but otherwise it looks good to me. Approved :)

elfjes (approved):

PR #142, opened 2026-06-26
Fix bug where input data may produce incorrectly shaped attribute arrays

elfjes:
Describe your changes
--------------------
When ``EntityInitDataFormat`` reads input data, we do not validate the shape of the input data arrays, which may lead to incorrect data arrays.

Furthermore, input data with only `None` values were always converted into 1D arrays which is not correct when we have a multidimensional attribute data type.

This PR fixes both these issues

Copyright license
-----------------

### DIFF
--- a/packages/simulation-core/src/movici_simulation_core/core/data_format.py
+++ b/packages/simulation-core/src/movici_simulation_core/core/data_format.py
@@ -87,11 +87,12 @@ class EntityInitDataFormat(ExternalSerializationStrategy):
 
             spec = self.schema.get_spec(name, default_data_type=infer_datatype, cache=True)
             data_type = spec.data_type
-
-            return t.cast(dict, parse_list(attr_data, data_type))
-
+            try:
+                return t.cast(dict, parse_list(attr_data, data_type))
+            except ValueError as e:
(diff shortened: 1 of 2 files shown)


onlyname0 (approved):
Looks good — building optimistically and then checking the shape is a clean way to handle this. I confirmed the multidim float `[None, None]` case now comes out as `(2, 2)` instead of collapsing to `(2,)`, and mis-shaped input raises like you'd expect. Happy to approve.

A few small, non-blocking things if you feel like it:

- The retry `np.array` on line 180 isn't wrapped, so genuinely ragged or non-numeric input skips your nice `ValueError` and leaks a raw numpy error instead (e.g. `[[1, 2], [3]]`). Might be worth catching that too so it all funnels through the same message.
- When the shape check does fire, the error doesn't say *which* attribute — `load_attribute` doesn't wrap `parse_list`, so a bad dataset just gets "input data is not of the correct shape" with no name. The `infer_datatype` path already does the `"...for '{name}'"` thing, so it'd be nice to match that.
- Tiny test nit: `test_create_array_raises_on_shape_mismatch` uses a bare `pytest.raises(ValueError)` — adding `match=` would stop it passing on an unrelated numpy error.

None of this blocks merging.

PR #143, opened 2026-07-08
Movici Data Core: API layer

elfjes:
Describe your changes
--------------------
This PR is the third and last of the three foundational PRs for movici data core. It introduces fastapi as the api layer. It is another quite large PR, but i've tried as much as possible to have the commits be meaningful self-contained pieces of work to try and make the development flow clear

Features
 - Endpoints
   - [x] Workspace
   - [x] Datasets
   - [x] Scenarios
   - [x] Updates
   - [x] Dataset types
   - [x] Entity types
   - [x] Attribute Types
   - [x] Model types
   - [x] Summary endpoints (for dataset and scenario)
 - [x] validation using pydantic
 - [x] Integration with docs

There is still more work to do on movici-data-core but this pr finishes the foundation

Copyright license
-----------------


packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -0,0 +1,64 @@
...
+
+@update_router.get("/")
+async def get_updates(backend: DepScenarioBackend) -> UpdateListOut:
+    scenarios = await backend.updates.list()
+    return UpdateListOut.from_domain(scenarios)
+
+
+@update_router.post("/")
+async def create_update(
+    backend: DepScenarioBackend, request: Request, filetype: DepAcceptFileType
```
onlyname0: **`POST /updates` decodes the request *body* using the `Accept` header instead of `Content-Type`.** `filetype: DepAcceptFileType` resolves from `request_filetype(accept=Header())` and is passed straight into `store_request_stream_to_disk` and `store_update_from_file`, which deserialize the posted bytes. Per RFC 9110, `Accept` states the desired *response* media type; the body's format is `Content-Type`.

Reproduced: a valid msgpack body sent with `Content-Type: application/x-msgpack` and the default `Accept: */*` → **HTTP 400** `invalid_data` (the bytes are parsed as JSON), while the byte-identical payload as JSON succeeds. `Content-Type` is never consulted on this path. The sibling `POST /datasets/{id}/data` gets this right via `UploadFile.content_type` (datasets.py:77).

Fix: infer the request filetype from `Content-Type`; keep `Accept` for response negotiation only (it is correctly used that way in `GET /updates/{id}`).

packages/data-core/src/movici_data_core/database/db_types.py
```diff
@@ -84,3 +85,25 @@ def process_result_value(self, value, dialect):
...
+        self.pattern = re.compile(pattern) if isinstance(pattern, str) else pattern
+        super().__init__(*args, **kwargs)
+
+    def process_bind_param(self, value, dialect):
+        if value is None:
+            return value
+
+        if not isinstance(value, str):
+            raise TypeError("must be a str")
+        if not re.match(self.pattern, value):
```
onlyname0: **Name validation is unanchored, producing both false 500s and silently accepted invalid names.** The pydantic side (`NameStr`/`AttributeNameStr`/`enum_name` in marshalling.py:68) uses `Field(pattern=...)`, which is *search* semantics, and this DB side uses `re.match`, which is *prefix-only*. Neither anchors `snake_case_pattern` (model.py:27) at the end, and the two disagree with each other.

Reproduced end-to-end:
- `POST /workspaces {"name":"Xa"}` → **HTTP 500**: pydantic matches the trailing `a`, then `re.match` rejects the leading `X`, raising `ValueError` from `process_bind_param`, which SQLAlchemy wraps as `StatementError` — not a `MoviciDataError`, so it bypasses the handler.
- `POST /workspaces {"name":"a-b"}` and `{"name":"my name!"}` → **HTTP 200**, persisting non-snake_case names.

The existing `test_snake_case` only tries `"A"`, which happens to fail in both layers, so this is untested.

Fix: anchor both layers — `re.fullmatch` here and `^...$` in the pydantic patterns (including the `enum_name` `[a-z][a-z_]*`). Separately, if this DB-layer check is meant as a second line of defense it should raise `MoviciValidationError` (422) like the length checks do, not a raw `ValueError` (500).

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -194,6 +222,135 @@ async def update(self, obj: Scenario, validator: ModelConfigValidator):
...
+            attribute_type = attribute.attribute_type
+            attribute_name = attribute_type.name
+            entity_group_name = attribute.entity_type.name
+
+            if attribute.attribute_type.name == "id" and is_dataset:
+                entity_counts[entity_group_name] = attribute.length
+
+            entity_summary = attribute_summaries.setdefault(entity_group_name, {})
+            if (attribute_summary := entity_summary.get(attribute_name)) is not None:
+                attribute_summary.min_val = min(attribute_summary.min_val, min_val)
```
onlyname0: **`min(None, ...)` / `max(None, ...)` raises `TypeError` → HTTP 500 in the scenario summary.** `DataType.get_min_max` returns `(None, None)` for `str`-typed attributes and for all-undefined chunks, and `DataArray.min_val/max_val` are nullable. This merge branch fires whenever the same `(entity_group, attribute)` pair appears more than once across the dataset query and the updates query — the normal case, since updates re-publish dataset attributes and multi-timestep scenarios repeat attributes across updates.

Reproduced: a `str` attribute present in the dataset plus one update makes `GET /scenarios/{id}/summary` fail with `TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'`, which escapes as an unhandled 500. The initial-creation branch tolerates `None` correctly; only this merge branch is broken.

Fix: guard both sides for `None` before merging, and add a summary test covering a str attribute updated within a scenario.

packages/data-core/src/movici_data_core/file_helpers.py
```diff
@@ -0,0 +1,103 @@
...
+    async def _store_to_disk():
+        with tempfile_delete_on_error(
+            suffix=filetype.default_extension, prefix=prefix, dir=path
+        ) as destination_file:
+            async for chunk in request.stream():
+                destination_file.write(chunk)
+        return pathlib.Path(destination_file.name)
+
+    loop = asyncio.get_running_loop()
+    return await loop.run_in_executor(executor, functools.partial(asyncio.run, _store_to_disk()))
```
onlyname0: **Streaming the request body via `asyncio.run()` in a worker thread hangs the endpoint for any non-trivial upload.** `_store_to_disk` iterates `request.stream()` (bound to the server's event loop) but this line runs it inside a *fresh* loop via `asyncio.run` on a `ThreadPoolExecutor` thread. uvicorn's `receive` awaits an `asyncio.Event` that binds to whichever loop first awaits it (the worker loop), while `.set()` fires on the server loop via a plain `call_soon` — the worker loop's selector is never woken, so `stream()` blocks forever.

Reproduced against a real uvicorn 0.46.0 server (not TestClient): a 1 KB `POST /updates` succeeds, but 128 KiB+ (past uvicorn's 64 KiB flow-control watermark) hangs until client timeout. Worse, each hung request permanently consumes one of the 10 shared executor threads; after ~10 such requests the entire upload surface (datasets included) is bricked until restart, with nothing logged. Tests pass only because TestClient and small bodies arrive fully buffered before `receive()` is awaited.

Fix: iterate `async for chunk in request.stream()` on the native loop and offload only the blocking `file.write(chunk)` (e.g. `run_in_executor` / `anyio.to_thread`).

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,93 @@
...
+@dataset_router.post("/")
+async def create_dataset(
+    dataset: ShortDatasetIn, backend: DepWorkspaceBackend
+) -> OperationSuccess:
+    result = await backend.datasets.create(dataset.to_domain())
+    return OperationSuccess(resource="dataset", id=result, verb="created")
+
+
+@dataset_router.get("/{dataset_id}")
+async def get_dataset(dataset_id: UUID, backend: DepBackend) -> ShortDatasetOut:
```
onlyname0, line 44: **Id-addressed routes are not workspace-scoped — cross-workspace read/mutate/delete by UUID.** `GET/PUT/DELETE /datasets/{id}`, `/datasets/{id}/data`, `/datasets/{id}/summary`, `GET/PUT/DELETE /scenarios/{id}` and `GET /updates/{id}` all take the unscoped `DepBackend`, and the underlying lookups filter on id alone (`DatasetRepository.get_by_id`/`delete`, `ScenarioRepository.get`/`delete`, `UpdateRepository.get_by_id`). Only `list`/`create` go through `DepWorkspaceBackend`, which is where the `# TODO: Validate authorization` hook lives.

Verified in MULTIPLE_WORKSPACES mode: a dataset created in workspace A is fetched, renamed and deleted with no `?workspace=` parameter at all. So the workspace binding enforced for list/create provides no isolation for any id-addressed operation, and the planned authorization will be silently bypassed by all of these routes once added.

Fix: route id-based endpoints through the workspace-bound backend and add `workspace_id == :workspace_id` to the `get_by_id`/`update`/`delete` predicates (and verify `Update → Scenario → workspace` for `GET /updates/{id}`).

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,93 @@
...
+@dataset_router.put("/{dataset_id}")
+async def update_dataset(
+    dataset_id: UUID, dataset: ShortDatasetIn, backend: DepBackend
+) -> OperationSuccess:
+    await backend.datasets.update(dataset_id, dataset.to_domain())
+    return OperationSuccess(resource="dataset", id=dataset_id, verb="updated")
+
+
+@dataset_router.delete("/{dataset_id}")
+async def delete_dataset(dataset_id: UUID, backend: DepBackend) -> OperationSuccess:
```
onlyname0, line 60: **Deleting an in-use, FK-referenced resource returns HTTP 500 (raw `IntegrityError`).** The schema uses `ondelete="RESTRICT"` for dataset_type/entity_type/attribute_type/model_type references and for `ScenarioDataset.dataset_id`/`Update.dataset_id`, but the delete methods (`GenericResourceRepository.delete`, `DatasetRepository.delete`) carry no `map_errors`, and the app-level handler only catches `MoviciDataError`.

Verified: `DELETE /datasets/{id}` for a dataset used by a scenario, and `DELETE /attribute_types|model_types|dataset_types/{id}` while referenced, all return 500 `Internal Server Error` (FOREIGN KEY constraint failed). Note `DatasetRepository.delete` also deletes the dataset's *data* before the failing row delete — only the per-request rollback prevents actual data loss.

Fix: wrap the delete methods with `map_errors` translating `IntegrityError` into a 409-style domain error ("resource is still referenced by …"), or pre-check references and raise a domain error.

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,93 @@
...
+async def get_dataset_data(dataset_id: UUID, backend: DepBackend):
+    # TODO: use accept header to serve msgpack or json
+    result = await backend.datasets.get_dataset_as_file(dataset_id)
+    if result is None:
+        raise ResourceDoesNotExist("dataset", id=dataset_id)
+    return FileResponse(result)
+
+
+@dataset_router.post("/{dataset_id}/data")
+async def create_dataset_data(dataset_id: UUID, data: fastapi.UploadFile, backend: DepBackend):
```
onlyname0: **Upload and download payloads are read, parsed and serialized synchronously on the event loop.** `create_dataset_data` carefully offloads the raw disk *write* to the executor, then immediately calls `update_from_file`, which runs `path.read_bytes()` + `serializer.loads(...)` + full numpy conversion + bounding-box computation directly on the loop (services/dataset.py:164); the update path does the same via `UpdateIn.read_from_file`. Downloads mirror it: `get_dataset_as_file`/`get_update_as_file` run `serializer.dumps(...)` over the whole payload and a synchronous write on the loop before `FileResponse` streams the result.

For multi-hundred-MB datasets this freezes every concurrent request for the full parse/serialize duration — offloading only the cheap write step buys nothing.

Fix: run the read+parse and serialize+write stages in the existing executor (`run_in_executor`/`anyio.to_thread`), keeping only the async DB calls on the loop. Related: `RawDataProcessor` reads binary uploads in blocking 100 MB `file.read` chunks and does blocking `orjson` encode/decode of full payloads on the loop.

packages/data-core/src/movici_data_core/api/dependencies.py
```diff
@@ -0,0 +1,65 @@
...
+
+SQLALCHEMY_SERVER_KEY = "__movici_sqlalchemy_server__"
+
+
+def server(request: Request) -> SQLAlchemyServer:
+    return getattr(request.state, SQLALCHEMY_SERVER_KEY)
+
+
+async def get_backend(server: DepServer):
+    async with server.get_backend() as backend:
```
onlyname0, line 20: **`COMMIT` runs in dependency teardown, after the success response has been sent.** `get_backend` yields the backend, and `SQLAlchemyServer.get_backend` calls `await session.commit()` in the context-manager exit — which, for a yield-dependency, executes *after* the response is produced and cannot alter it. So every mutating endpoint sends its `OperationSuccess` ("created"/"updated", including the new id) before `COMMIT` runs.

If that commit then fails — SQLite `database is locked` under the concurrent-writer setup `make_default_app` builds (file-backed DB, no busy_timeout/WAL), or any deferred constraint — the transaction rolls back, nothing persists, the `MoviciDataError` handler never runs, and the client already holds a 200 referencing a row that does not exist. Serial TestClient tests cannot surface this.

Fix: commit inside the request scope before the response is built (an explicit `await backend.commit()` at the end of mutating handlers, or a middleware that commits before forwarding response-start), leaving teardown to rollback/close only.

packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -0,0 +1,64 @@
...
+    UpdateWithDataOut,
+)
+from movici_simulation_core.types import FileType
+
+from .dependencies import DepBackend, DepScenarioBackend
+
+update_router = APIRouter(prefix="/updates")
+
+
+def request_filetype(accept: t.Annotated[str | None, Header()] = None) -> FileType:
```
onlyname0: **`Accept` is treated as a single MIME type, so any conformant client gets a 415 on both update endpoints.** `request_filetype` special-cases only the exact string `*/*`; anything else goes to `infer_filetype_from_filename_or_mimetype`, whose `base_mimetype` (file_helpers.py:90) does just `content_type.split(";")[0].strip()` — it never splits on `,` and never handles q-values.

Verified against the app:
```
request_filetype('application/json, text/plain, */*')                        -> FileType.OTHER   # axios default
request_filetype('text/html,application/xhtml+xml,...;q=0.9,*/*;q=0.8')      -> FileType.OTHER   # browser default
GET  /updates/{id}  accept: 'application/json, */*'  -> 415 {"message":"Filetype .dat is not supported..."}
POST /updates       accept: 'application/json, */*'  -> 415 (same)
```
Only clients sending exactly `*/*` (the httpx/requests default — which is why the suite is green) or exactly one MIME type work.

This is a separate root cause from the `Accept`-vs-`Content-Type` issue on POST: it also breaks `GET /updates/{id}`, where consulting `Accept` is the *correct* design.

Fix: parse `Accept` properly — split on `,`, strip `;q=` params, take the first entry present in `MIMETYPES`, and fall back to `FileType.JSON` for `*/*`/no match instead of 415-ing.

packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -0,0 +1,64 @@
...
+async def get_updates(backend: DepScenarioBackend) -> UpdateListOut:
+    scenarios = await backend.updates.list()
+    return UpdateListOut.from_domain(scenarios)
+
+
+@update_router.post("/")
+async def create_update(
+    backend: DepScenarioBackend, request: Request, filetype: DepAcceptFileType
+) -> OperationSuccess:
+    tempfile = await store_request_stream_to_disk(request, backend.tmpfile_dir, filetype=filetype)
```
onlyname0, line 48: **A duplicate `POST /updates` (same `scenario_id, timestamp, iteration`) returns HTTP 500 instead of 409.** `Update` carries `UniqueConstraint("scenario_id", "timestamp", "iteration")` (model.py:520), but `UpdateRepository.create` is the only create method in this PR without a `@map_errors({IntegrityError: ResourceAlreadyExists})` decorator, and nothing on the `store_update_from_file` path handles it. Verified: a second update with the same key raises a raw `IntegrityError`, which the `MoviciDataError` handler does not catch → 500.

A simulation client re-submitting a result (an ordinary retry) hits this, and gets a 500 where every sibling resource returns a clean 409.

Fix: decorate `UpdateRepository.create` with `map_errors` translating `IntegrityError` into `ResourceAlreadyExists("update", ...)` including timestamp/iteration in the message.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -131,6 +144,13 @@ async def delete(self):
         id = self._ensure_scenario_id()
         await self.session.execute(delete(db.Scenario).where(db.Scenario.id == id))
```
onlyname0, line 177: **Deleting a scenario (or a workspace) orphans the `Attribute`/`DataArray` blob rows.** `ScenarioRepository.delete()` executes only `delete(Scenario)`. FK cascades remove `Update` and `UpdateAttribute` link rows, but the `Attribute` rows and their `DataArray` blobs referenced *by* `UpdateAttribute` are never deleted (that cascade runs when the `Attribute` is deleted, not the reverse). `UpdateRepository.delete_all` already implements the needed cleanup — delete `Attribute` rows first — but delete does not call it; `WorkspaceRepository` uses the generic delete and likewise skips dataset-data cleanup.

Verified with FK enforcement on: deleting a scenario left 2 orphaned `attribute`/`data_array` rows; deleting the whole workspace left 6 orphans with zero workspaces/datasets/scenarios/updates remaining. These are the tables holding the binary array data, so the DB grows unboundedly with every deletion.

Fix: run the `delete_all` attribute cleanup before deleting the scenario, and delete dataset data plus update attributes for all contained resources before deleting a workspace.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -200,17 +237,29 @@ def from_domain(cls, obj: ScenarioModel):
...
 
 
-class ScenarioOut(OutModel[Scenario]):
+class ShortScenarioOut(OutModel[Scenario]):
     id: UUID
     name: str
     display_name: str
+    created_at: datetime.datetime
+    updated_at: datetime.datetime
+    status: ScenarioStatus
```
onlyname0, line 243: **`status` can never reflect the stored value — every scenario reports `"ready"`.** The `scenario.status` column exists (`status: Mapped[ScenarioStatus]`, model.py:409) but `db.Scenario.to_domain()` doesn't pass it, so the domain default `ScenarioStatus.READY` (domain_model.py:248) is what this new output field always serializes — on both the list and detail endpoints, regardless of whether the row says `RUNNING` or `FAILED`.

I realize `ScenarioStatus` handling is explicitly flagged as unfinished (the TODO at domain_model.py:34 and on `set_status`), but this PR newly *exposes* the field over the API, so it currently ships a field that looks authoritative and is a constant. Either populate it in `to_domain()` or leave it out of the Out models until the status lifecycle is implemented.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -200,17 +237,29 @@ def from_domain(cls, obj: ScenarioModel):
...
+    has_updates: bool
```
onlyname0, line 244: **`has_updates` is always `false` on the list endpoint.** It's only populated in `ScenarioService.get()` (services/scenario.py:29, via `updates.exists()`); `ScenarioService.list()` returns `repository.scenarios.list()` untouched and `db.Scenario.to_domain()` never sets it, so the dataclass default `False` is what gets serialized.

Verified: after creating a scenario and POSTing one update, `GET /scenarios?workspace=…` reports `"has_updates": false` while `GET /scenarios/{id}` on that same scenario reports `true`. Any UI deciding "does this scenario have results / can it be reset" from the list gets the wrong answer for every row.

Contrast `ShortDatasetOut.has_data`, which `DatasetRepository.list()` correctly computes with an `exists()` subquery — the same treatment would fix this (add `exists().where(db.Update.scenario_id == db.Scenario.id)` to `ScenarioRepository.list()`).

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -423,3 +477,200 @@ def to_domain(self):
...
+class ModelTypeOut(OutModel[ModelType]):
+    id: UUID
+    name: str
+    jsonschema: dict
+
+
+class DatasetSummaryOut(OutModel[DatasetSummary]):
+    general: dict
+    epsg_code: int | None
+    bounding_box: BoundingBoxField
```
onlyname0, line 641: **`bounding_box` serializes to `null` but is documented as a required, non-nullable 4-number array.** `BoundingBoxField` pairs `PlainSerializer(lambda bbox: bbox.as_tuple_or_none())` — which returns `None` when any coordinate is `None` — with a `WithJsonSchema({"type":"array","minItems":4,"maxItems":4,…})` that has no null branch. Since this field is non-optional it lands in `required`.

Verified: `GET /scenarios/{id}/summary` for a dataset with no geometry returns `{"general":{},"epsg_code":null,"bounding_box":null,…}` while `components.schemas.DatasetSummaryOut` requires a 4-item number array. Non-spatial datasets (tabular, plain entity data) are the common case, so strict/generated clients break on both summary endpoints.

Fix: make the schema match the serializer — `WithJsonSchema({"anyOf": [{"type":"array","minItems":4,"maxItems":4,"items":{"type":"number"}}, {"type":"null"}]})` — and/or declare `bounding_box: BoundingBoxField | None`.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -423,3 +477,200 @@ def to_domain(self):
...
+    min_val: bool | int | float | None
+    max_val: bool | int | float | None
+
+
+class OperationSuccess(BaseModel):
+    resource: str
+    id: UUID | str
+    verb: str
+
+    @model_serializer
```
onlyname0: **This `@model_serializer` makes the generated OpenAPI response schema wrong for every mutating endpoint.** `OperationSuccess` declares `resource`/`id`/`verb`, but `serialize()` emits `{"result", "id", "message"}`. FastAPI derives the response schema from the declared fields and ignores an untyped model serializer, so `app.openapi()` documents:

```
components.schemas.OperationSuccess = {properties: {resource, id, verb},
                                      required: ["resource", "id", "verb"]}
```

while the wire body is `{"result":"ok","id":"…","message":"scenario created"}` (as your own tests assert). Two of the three documented *required* properties never appear in any response. This affects all ~20 POST/PUT/DELETE operations across `schema.py`, `datasets.py`, `scenarios.py`, `updates.py`, `workspaces.py`, so any generated client, contract test, or published doc breaks on every write. Runtime doesn't catch it because FastAPI validates the `OperationSuccess` instance by isinstance pass-through and only then applies the serializer.

Fix: declare the real output fields (`result`/`id`/`message`, e.g. via computed fields) or give `serialize()` a typed return, so the declared schema matches what is emitted.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -59,10 +58,15 @@ async def create(self, dataset: Dataset):
...
+
+    async def get_dataset_as_file(
+        self, dataset_id: UUID, filetype: FileType = FileType.JSON
+    ) -> pathlib.Path | None:
         existing = await self.repository.datasets.get_by_id(dataset_id)
         if existing is None:
-            raise ResourceDoesNotExist("dataset", id=dataset_id)
+            return None
 
         with tempfile_delete_on_error(
```
onlyname0, line 71: **Temp files are never cleaned up — unbounded growth in `tmpfile_dir`.** A grep of `packages/data-core/src` finds no `unlink`/`BackgroundTask`/`background=` outside the exception-only cleanup in `tempfile_delete_on_error` (which creates files with `delete=False`, so success paths always leave the file). Download temp files from `get_dataset_as_file`/`get_update_as_file` are served via `FileResponse(...)` with no background task, and starlette's `FileResponse` never deletes the file it serves. Upload staging files from `store_file_to_disk`/`store_request_stream_to_disk` are read and then left behind. `make_default_app`'s `mkdtemp` directory is never removed either.

Every successful `GET/POST /datasets/{id}/data` and `GET/POST /updates` therefore leaks one file.

Fix: `FileResponse(path, background=BackgroundTask(os.unlink, path))` for downloads, delete staged uploads after ingestion, and clean the mkdtemp directory on shutdown.

packages/data-core/tests/test_file_helpers.py
```diff
@@ -0,0 +1,111 @@
...
+    with (
+        pytest.raises(RuntimeError),
+        tempfile_delete_on_error(suffix=".suff", prefix="prefpref", dir=tmp_path) as file,
+    ):
+        file.write(b"somedata")
+        raise RuntimeError
+    assert not pathlib.Path(file.name).exists()
+
+
+async def test_store_file_to_disk(tmp_path):
```
onlyname0, line 43: **`store_request_stream_to_disk` has zero direct coverage, and the streaming path is structurally unreachable from `TestClient`.** This file tests `store_file_to_disk` thoroughly but never imports or calls `store_request_stream_to_disk` (file_helpers.py:52) — the sole ingestion path for `POST /updates`.

The only indirect coverage goes through `TestClient`, which hands the ASGI app the entire body as one in-memory chunk, so `async for chunk in request.stream()` always iterates exactly once and the `asyncio.run`-in-executor construct never has to interact with the outer loop. Confirmed: a ~1 MB / 20k-entity `POST /updates` through the existing fixture passes in 0.6 s. That is precisely why the deadlock in `file_helpers.py:70` is invisible to CI.

Fix: add a direct unit test with a stub `Request` whose `.stream()` is an async generator yielding several chunks (asserting contents and no deadlock), and/or exercise at least one upload against a live uvicorn instance rather than `TestClient`.

packages/data-core/pyproject.toml
```diff
@@ -29,17 +29,23 @@ classifiers = [
...
-sqlite = ["aiosqlite>=0.22.1"]
 
 [dependency-groups]
-dev = ["pytest-asyncio>=1.3.0"]
+dev = [
+  "httpx>=0.23.0,<1.0.0",      # from fastapi[standard]
+  "pytest-asyncio>=1.3.0",
+  "uvicorn[standard]>=0.12.0", # from fastapi[standard]
+]
+sqlite = ["aiosqlite>=0.22.1"]
```
onlyname0: The `sqlite` extra moved from `[project.optional-dependencies]` to `[dependency-groups]` (PEP 735). Dependency groups are **not** published as extras — I built the wheel from this branch and its METADATA contains no `Provides-Extra` and no `aiosqlite` `Requires-Dist` at all, so `pip install movici-data-core[sqlite]` no longer installs anything. Meanwhile `make_default_app` hardcodes `sqlite+aiosqlite:///...`, so end users need `aiosqlite` at runtime but can no longer express it via the extra. If uv-only workflows are the intent that's fine — otherwise keep `sqlite` as a real optional-dependency extra.

packages/data-core/src/movici_data_core/api/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+    for router in routers:
+        app.include_router(router)
+    return app
+
+
+def make_default_app():
+    tmpfile_path = pathlib.Path(tempfile.mkdtemp(prefix="movici_api_tmp_"))
+
+    db_path = pathlib.Path("movici.db`")
```
onlyname0: Cluster of issues in `make_default_app` / `__main__`:
1. `pathlib.Path("movici.db`")` — stray backtick, so the database file is literally named ``movici.db` ``.
2. `asyncio.run(setup())` raises `RuntimeError: asyncio.run() cannot be called from a running event loop` when the app is served as `uvicorn movici_data_core.api:make_default_app --factory` (uvicorn calls factories inside its already-running loop). Because of the backtick the `db_path.exists()` guard is effectively always false, so this path runs on every startup. Move the DB setup into the `lifespan`, which is already async.
3. The `__main__` block builds the app but never starts a server, so there is no working entrypoint.
4. `logger.exception("An error occured")` in the handler above — typo ("occurred").

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,93 @@
...
+dataset_router = fastapi.APIRouter(prefix="/datasets")
+
+
+@dataset_router.get("/")
+async def get_datasets(backend: DepWorkspaceBackend) -> DatasetListOut:
+    datasets = await backend.datasets.list()
+    return DatasetListOut.from_domain(datasets)
+
+
+@dataset_router.post("/")
```
onlyname0: Collection routes are registered only as `"/"` under the router prefix, so the canonical URLs (`/datasets`, `/scenarios`, `/workspaces`, the four type routers) only work via a 307 redirect. For `GET` that's just a wasted round-trip, but a client that doesn't re-send the body/headers on redirect (or one following 307 → 308 semantics inconsistently) will fail on `POST`/`PUT`. Your own tests use both spellings (`"/datasets"` and `"/updates/"`), so the redirect is already in play. Registering the routes as `""` instead of `"/"` serves the canonical path directly.

packages/data-core/src/movici_data_core/api/scenarios.py
```diff
@@ -0,0 +1,100 @@
...
+    ScenarioOut,
+)
+from movici_data_core.validators import ModelConfigValidator
+
+from .dependencies import DepBackend, DepWorkspaceBackend, allow_in_modes
+
+scenario_router = APIRouter(prefix="/scenarios")
+
+
+async def model_config_validator(backend: DepBackend):
```
onlyname0, line 22: `asyncio.gather(backend.attribute_types.list(), backend.entity_types.list())` runs two queries concurrently on a **single `AsyncSession`**. SQLAlchemy explicitly documents `AsyncSession` as unsafe for concurrent use in multiple tasks; overlapping operations can raise `InvalidRequestError` ("this session is provisioning a new connection; concurrent operations are not permitted") or corrupt session state. It happens to survive today because both statements are short and a connection is already checked out, but it's a latent failure that will surface under load or a slower dialect. Just `await` them sequentially — there's no real parallelism to gain from one connection.

packages/data-core/src/movici_data_core/api/scenarios.py
```diff
@@ -0,0 +1,100 @@
...
+@scenario_router.put("/{scenario_id}")
+async def update_scenario(
+    scenario_id: UUID,
+    scenario: ScenarioIn,
+    backend: DepBackend,
+    validator: DepModelConfigValidator,
+) -> OperationSuccess:
+
+    await backend.for_scenario(scenario_id).scenarios.update(
+        dataclasses.replace(scenario.to_domain()), validator
```
onlyname0: `dataclasses.replace(scenario.to_domain())` with no field overrides is a no-op copy — pass `scenario.to_domain()` directly.

packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@update_router.delete("/")
+async def delete_updates(backend: DepScenarioBackend) -> OperationSuccess:
+    await backend.updates.delete_all()
+    assert backend.scenario_id is not None
+    return OperationSuccess(resource="updates", id=backend.scenario_id, verb="deleted")
+
+
+@update_router.get("/{update_id}")
```
onlyname0, line 63: `get_update` is annotated `-> UpdateWithDataOut` but actually returns a `FileResponse` (via `t.cast`). So OpenAPI advertises a JSON object schema for an endpoint that streams a file whose media type may be msgpack. The `t.cast` silences the type checker without changing the contract. Consider declaring `response_class=FileResponse` and `responses={200: {"content": {...}}}` (or `response_model=None`) so the published schema reflects what's served.

packages/data-core/src/movici_data_core/database/general.py
```diff
@@ -128,4 +128,10 @@ def _default_flags(mode: DatabaseMode):
             "STRICT_MODEL_TYPES": True,
             "STRICT_SCENARIO_DATASETS": True,
         }
-    return {}
+    return {
```
onlyname0, line 131: `_default_flags` is applied only at `initialize_database` (general.py:63). `SQLAlchemyBackend.set_database_mode` changes the mode but never re-applies the flags, so a database upgraded `SINGLE_SCENARIO → MULTIPLE_WORKSPACES` keeps the permissive `STRICT_* = False` values, while a freshly initialized MULTIPLE_WORKSPACES database gets `STRICT_* = True`. Two databases in the same mode end up with different validation behavior.

Worth noting the new test fixture papers over exactly this: `tests/conftest.py:138` calls `backend.set_options(**_default_flags(database_mode))` after `set_database_mode`, so the tests never exercise the production path. Either re-apply the defaults inside `set_database_mode` or make the flag semantics on upgrade explicit.

packages/data-core/src/movici_data_core/database/repository/general.py
```diff
@@ -27,6 +30,13 @@ class DatasetTypeRepository(GenericResourceRepository[DatasetType]):
     __resource__ = db.DatasetType
     __resource_type_name__ = "dataset_type"
 
+    @map_errors(
```
onlyname0, line 33: The blanket `IntegrityError → ResourceAlreadyExists(409)` mapping also mislabels FK and NOT NULL violations as "duplicate". At these type repositories it happens to be safe (no FK columns, and the In-models supply every NOT NULL field), so only the unique-name case is reachable and it maps correctly. But at `DatasetRepository.create` and `ScenarioRepository.create`, which insert a `workspace_id` FK, a workspace deleted between dependency validation and insert yields a 409 "duplicate" instead of a 404 — a race, but a misleading one. Consider inspecting the constraint (`exc.orig` / constraint name) rather than assuming uniqueness.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -194,6 +222,135 @@ async def update(self, obj: Scenario, validator: ModelConfigValidator):
...
+
+        for (attribute, min_val, max_val), is_dataset in itertools.chain(
+            zip(attributes_from_dataset, itertools.repeat(True)),
+            zip(attributes_from_updates, itertools.repeat(False)),
+        ):
+            attribute_type = attribute.attribute_type
+            attribute_name = attribute_type.name
+            entity_group_name = attribute.entity_type.name
+
+            if attribute.attribute_type.name == "id" and is_dataset:
```
onlyname0, line 352: Question on intent: `entity_counts` is filled only for `id` attributes where `is_dataset` is true, so an entity group whose ids arrive **only via updates** is reported with `count: 0` while still being listed with all of its attributes and their min/max — and `count` (the total, line 351) omits it too.

Verified: a scenario whose dataset is a stub (exactly what `ensure_scenario_datasets` creates when `STRICT_SCENARIO_DATASETS` is unset — the default for the `SINGLE_SCENARIO` database `make_default_app()` builds) with all data arriving through updates returns `count=0, entity_groups=[('roads', 0)]` even though three entities exist.

I can see the argument that summary counts are deliberately dataset-derived (as the neighbouring bounding-box comment says), so flagging rather than asserting a bug — but the result is internally inconsistent: the group is present with populated attributes and a zero count. If it's intended, a comment would help; if not, fall back to the update-sourced `id` length when the dataset has none.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -213,9 +370,11 @@ async def _store_scenario_details(
         repository = self.all_data.for_workspace(workspace_id)
         scenario_datasets = []
         if obj.datasets:
+            # deduplicate datasets by name
+            datasets_by_name = {ds.name: ds for ds in obj.datasets}
```
onlyname0: The dedup keeps the **last** occurrence (`{ds.name: ds for ds in obj.datasets}`), so when a scenario config lists the same dataset twice with different types, the earlier explicitly-supplied type is silently discarded rather than reported as a conflict. It also renumbers the list, so any `MoviciValidationError` path raised downstream will carry an index that no longer matches the client's submitted `datasets[i]`. Consider rejecting genuine duplicates (or at least duplicates that disagree on type) instead of silently collapsing them.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -242,19 +401,26 @@ async def _store_scenario_details(
...
+                        "scenario_id": scenario_id,
+                        "model_type_id": model_type.id,
+                        "sequence": idx,
+                        "config": self._stripped_config(model),
+                    }
+                    for idx, (model, model_type) in enumerate(zip(scenario_models, model_types))
+                ],
+            )
+        except sqlalchemy.exc.IntegrityError as exc:
+            model_name: str = exc.params[1]  # type: ignore
```
onlyname0: `exc.params[1]` to recover the duplicate model name is fragile, and already returns the wrong value today. The `.returning()` executemany triggers SQLAlchemy `insertmanyvalues`, so `exc.params` is a single flattened tuple of *all* rows; `params[1]` is therefore always the **first** row's name, not the duplicate's — with names `[alpha, beta, beta]` it yields `'alpha'`. It also depends on the client-side uuid default occupying index 0 (column-order fragile) and on a positional paramstyle (a dict paramstyle such as psycopg2's would `KeyError`), while `db_types.py` branches on postgresql/mssql, so multi-dialect use is clearly intended. The existing test only passes because both models share the same name.

You already have the names locally — a Python-side `len({m.name for m in models}) != len(models)` check is deterministic, reports the right name, and drops the `# type: ignore`.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -423,3 +477,200 @@ def to_domain(self):
...
+
+class ModelTypeIn(InModel[ModelType]):
+    name: NameStr
+    jsonschema: dict
+
+    @field_validator("jsonschema", mode="after")
+    @classmethod
+    def _validate_jsonschema(cls, schema):
+        try:
+            movici_validator(schema).check_schema(schema)
```
onlyname0, line 623: `_validate_jsonschema` only runs `check_schema`, which confirms a `pattern` is a *syntactically* valid regex but says nothing about catastrophic backtracking. A client can `POST /model_types` a schema containing `"pattern": "^(a+)+$"`; it is stored, and later scenario create/update validates model configs against it, applying that regex to user-supplied strings — synchronously, on the event loop. Verified in-tree: `check_schema` accepts it and `iter_errors` against `"a"*40+"b"` does not return within 6s, pegging a core and stalling every other request (compounded by the acknowledged no-auth TODO). Consider linting submitted patterns (or disallowing `pattern`), and running schema validation off the event loop with a timeout.

packages/data-core/src/movici_data_core/services/common.py
```diff
@@ -57,3 +38,68 @@ async def update(self, id: UUID, obj: T_dom):
...
+
+    # try scenario as a UUID
+    try:
+        scenario_id = UUID(scenario_name_or_id)
+    except ValueError:
+        scenario_id = None
+
+    if scenario_id is not None:
+        # scenario was given as a uuid
+        scenario_obj = await repository.scenarios.for_id(scenario_id).get()
```
onlyname0, line 86: When the scenario is supplied as a UUID, `ensure_valid_scenario` ignores the `workspace` parameter entirely — `repository.scenarios.for_id(scenario_id).get()` filters only on `Scenario.id`, and `workspace_name_or_id` is never validated or compared. Verified: `GET /updates/?scenario=<uuid in ws_A>&workspace=ws_B` returns ws_A's updates even though the client explicitly scoped the request to a different, valid workspace; the same dependency backs `POST /updates` and `DELETE /updates/`, so writes and deletes are affected too. When both are supplied, resolve the workspace and verify `scenario.workspace_id` matches, raising `ResourceDoesNotExist` on mismatch.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -134,6 +158,7 @@ class Options(Base):
     STRICT_ATTRIBUTE_TYPES: Mapped[bool] = mapped_column(default=False)
     STRICT_MODEL_TYPES: Mapped[bool] = mapped_column(default=False)
     STRICT_SCENARIO_DATASETS: Mapped[bool] = mapped_column(default=False)
+    IMMUTABLE_WORKSPACE_NAMES: Mapped[bool] = mapped_column(default=False)
```
onlyname0, line 161: This adds an `Options` column (and a new `UniqueConstraint` on `scenario_dataset` at :438) while `DEFAULT_SCHEMA_VERSION` stays `"v1"` and there's no migration; `setup_db` uses `create_all`, which creates missing tables but never `ALTER`s an existing one, and `get_version` is never compared to anything outside tests. A database created before this PR then fails every request with `OperationalError: no such column: options.IMMUTABLE_WORKSPACE_NAMES`.

Entirely reasonable to defer while data-core is pre-alpha and unreleased (earlier commits on this branch changed the schema the same way), so flagging only to get it tracked: at minimum, compare `get_version()` against `DEFAULT_SCHEMA_VERSION` when opening a backend so the failure is a clear "schema outdated" error rather than a raw 500.

packages/data-core/tests/api/test_dependencies.py
```diff
@@ -0,0 +1,91 @@
...
+    "database_mode, scenario_q, workspace_q, error",
+    [
+        (DatabaseMode.MULTIPLE_WORKSPACES, None, None, InvalidAction),
+        (DatabaseMode.SINGLE_WORKSPACE, None, A_WORKSPACE_NAME, InvalidAction),
+        (DatabaseMode.MULTIPLE_WORKSPACES, None, A_WORKSPACE_ID, InvalidAction),
+        (DatabaseMode.MULTIPLE_WORKSPACES, A_SCENARIO_NAME, None, InvalidAction),
+        (DatabaseMode.MULTIPLE_WORKSPACES, A_SCENARIO_NAME, "asdf", ResourceDoesNotExist),
+        (DatabaseMode.MULTIPLE_WORKSPACES, "asdf", A_WORKSPACE_NAME, ResourceDoesNotExist),
+        (DatabaseMode.MULTIPLE_WORKSPACES, UUID(int=0), A_WORKSPACE_ID, ResourceDoesNotExist),
+        (DatabaseMode.MULTIPLE_WORKSPACES, A_SCENARIO_NAME, "asdf", ResourceDoesNotExist),
```
onlyname0: Duplicate parametrize row: `(DatabaseMode.MULTIPLE_WORKSPACES, A_SCENARIO_NAME, "asdf", ResourceDoesNotExist)` already appears at line 67.

packages/data-core/tests/api/test_scenarios.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    locs = [".".join(str(f) for f in e["loc"]) for e in result["detail"]]
+    assert "models.0.type" in locs[0]
+    assert "models.0.type" in locs[1]  # a second message at this paht for the union type
+    assert "datasets.0.name" in locs[2]
+    assert "datasets.0.type" in locs[3]
+    assert "datasets.0.type" in locs[4]  # a second message at this paht for the union type
+
+
+async def _prepare_and_expected_summary(
```
onlyname0: `_prepare_and_expected_summary` is defined but never called — `test_get_summary_by_dataset_id` duplicates its entire body inline. Drop one of the two.

packages/data-core/src/movici_data_core/file_helpers.py
elfjes: hmm good point. Running every chunk write in a thread might not be a very performant option, but it might be the best we have

packages/data-core/src/movici_data_core/api/updates.py
elfjes: omg sometimes I'm such an idiot. Thanks!

packages/data-core/src/movici_data_core/database/db_types.py
elfjes: nice catch, thanks

as for the exception type, I'm a bit flexible with the exception types coming from the database model. Maybe they should all be low level types and we translate them to other types in the Repository. However, I would like to leave that until later. The main consumer of the repository is the service layer and by extension the api layer, and as long as we send proper errors (in this case validation error) to the user, I'm fine with that. Is that ok for you?

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: thanks for the catch :)

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: every uuid is unique and links to at most one object, which might be scoped, such as to a workspace. There is no leakage between workspaces because there is no overlap in object uuids between workspaces. So there is no need to enforce clients also specifying the workspace when they already supply a uuid.

When we implement authorization, we lookup the object by uuid and check its workspace. If the user has access to that workspace, we return the object, otherwise 403.

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: Good point, forgot to check for these restrictions, going for the map_errors approach, which works well, but the error messages can be more useful if we'd check for the objects beforehand. We can then also point to where exactly they are still used. But that's for later

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: Yes this is correct. Since this is cpu bound work, we must resort to multiprocessing in order to parallelize this (or subinterpreters or free-threaded python, both of which are only available in newer versions of python). However, doing this correctly and robust is hard. We have a production-ready solution in the data_engine code base on gitlab which I intent to reuse over there. However, due to legal reasons we cannot take it in here. It's also very nontrivial to reimplement. 

There are other implementations out there, such as any.io, but our own implementation has:
- process initialization so that it can connect to a database on initialization and keep a connection alive during its lifetime
- retry in case a process or processpool dies

I do see that we could build an implementation that uses any.io. But I would like to make that its own PR.

packages/data-core/src/movici_data_core/api/dependencies.py
elfjes: caught this by setting the dependencies to "function" scope, which makes sure they close before sending the responise. Hard to test though, since exception handling for the test client is different from running in production or through uvicorn

packages/data-core/src/movici_data_core/api/updates.py
elfjes: request_filetype was not meant for accept, but for content type. We're currently not handling any accept headers (after my fix with the accept/content-type confusion)

packages/data-core/src/movici_data_core/api/updates.py
elfjes: thanks for the catch :)

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: nice catch, but implement differently

packages/data-core/src/movici_data_core/marshalling.py
elfjes: Yep, scenario status still needs to be implemented properly. In fact there needs to be some coordination between the core scenario status (can it be run? <- this is mainly a question of do all associated datasets have data?) and the simulation status. The simulation status needs to be updated by the simulation during its run. I have ideas, but will implement them in a different PR

packages/data-core/src/movici_data_core/marshalling.py
elfjes: Thanks :)

packages/data-core/src/movici_data_core/marshalling.py
elfjes: thanks for the catch :)

packages/data-core/src/movici_data_core/marshalling.py
elfjes: good point :+1:

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: This goes back to the "when do we clean up files" discussion from earlier.

I'm not sure that it's save to remove the file in a background task. There might be a risk that the file is removed while it is still being used to serve out the response. So we need to be very careful. While important, I want to postpone this to a separate PR

packages/data-core/tests/test_file_helpers.py
elfjes: Not quite sure what you mean by deadlock, but sure, I  can add another test

packages/data-core/pyproject.toml
elfjes: good point, I'm adding a [standard] optional dependency group that installs the requirements to run the default app

packages/data-core/src/movici_data_core/api/__init__.py
elfjes: Thanks, I've refactored and moved the default app creation (required for fastapi-cli) in a ``main.py`` and added a barebones uvicorn start command. We might want to extend the `main.py` to a more proper cli application at some point, but that's out of the scope of this PR. Preferred way of running the app directly from the repo is using the fastapi cli. Well, actually the preferred way to treat movici_data_core as a library which requires you to implement your own app creation process, but hey, I guess we want to provide some sane defaults

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: Whether or not the canonical url for a collection endpoint should end with a trailing slash is up for debate. However, I've rewritten the urls so that none of them end with a slash

packages/data-core/src/movici_data_core/api/scenarios.py
elfjes: hmm I had not realized that, thanks :blush:

packages/data-core/src/movici_data_core/api/scenarios.py
elfjes: thnx for the catch

packages/data-core/src/movici_data_core/api/updates.py
elfjes: We don't yet properly support changing the reponse content-type based on the accept header. So for now we always return json. The fact that we're sending out a file response should be transparent to the consumer. It should just interpret it as a json (or later, msgpack) response. Which is what the response annotation is for. If we'd change that to fileresponse, the user would not know what is inside the file, even though it has a strict schema

packages/data-core/src/movici_data_core/database/general.py
elfjes: Yes, this is by design. The default options are just that, default. if the user upgrades (or downgrades) the database, it would be confusing if we also reset the flags.

In the tests, we assume that the db is freshly created, and therefore expect the default flags

packages/data-core/src/movici_data_core/database/repository/general.py
elfjes: Thanks :), I've added additional checks through the different usages of `map_errors`

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: creating new entities through updates is currently not yet supported throughout the whole stack. Some parts of the stack do support it, but not many. In a way it can be considered undefined behaviour. But I'll add a comment to clarify this

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: fair point

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: fair point, changed to checking for duplicates in advance

packages/data-core/src/movici_data_core/marshalling.py
elfjes: fair point, but not something for this PR. Currently only admins can upload jsonschemas or someone is running their own instance of the backend. But good to keep in mind for the future

packages/data-core/src/movici_data_core/services/common.py
elfjes: scenario uuid fully resolves to a unique scenario. Workspace is only there when supplying scenario name and may be safely ignored in this case. While I do see that this may lead to some confusion (in case someone implements an api client themselves, which I don't see happening very soon), we'd need to make these kinds of checks in many more places, so keeping it out of scope for now, until it becomes an actual issue.

packages/data-core/src/movici_data_core/database/model.py
elfjes: Will do this once we make changes while released

packages/data-core/tests/api/test_dependencies.py
elfjes: thanks

packages/data-core/tests/api/test_scenarios.py
elfjes: thanks for the catch :)

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,97 @@
...
+    return OperationSuccess.for_path_operation(resource="dataset", id=dataset_id, verb="deleted")
+
+
+@dataset_router.get("/{dataset_id}/data")
+async def get_dataset_data(dataset_id: UUID, backend: DepBackend):
+    # TODO: use accept header to serve msgpack or json
+    result = await backend.datasets.get_dataset_as_file(dataset_id)
+    if result is None:
+        raise ResourceDoesNotExist("dataset", id=dataset_id)
+    return FileResponse(result)
```
onlyname0: **Temp file leak.** `get_dataset_as_file` writes a serialized copy of the dataset into
`tmpfile_dir` and returns the path, but nothing ever deletes it — `tempfile_delete_on_error`
only unlinks on the *error* path, and the success path has no cleanup anywhere in the package
(the only `os.unlink` in `src/` is that error handler).

Three `GET /datasets/{id}/data` requests against the test app leave three files behind:

```
after 3 downloads: ['dataset-f215ac54-…-a_transport_network-e79nd828.json',
                    'dataset-f215ac54-…-a_transport_network-ocqe_ukv.json',
                    'dataset-f215ac54-…-a_transport_network-vledhh8y.json',
                    'movici.db']
```

Since these are full copies of the dataset, disk use grows with traffic and never comes back.
Starlette can clean up after the response is flushed:

```python
from starlette.background import BackgroundTask

return FileResponse(result, background=BackgroundTask(os.unlink, result))
```

Same fix needed on `updates.py:66`.

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,97 @@
...
+@dataset_router.post("/{dataset_id}/data")
+async def create_dataset_data(dataset_id: UUID, data: fastapi.UploadFile, backend: DepBackend):
+    if not await backend.datasets.get(id=dataset_id):
+        # We also check this in the DatasetService, but we short circuit here to prevent additional
+        # work with handling the incoming file
+        raise ResourceDoesNotExist("dataset", id=dataset_id)
+
+    mimetype = base_mimetype(data.content_type)
+    filetype = infer_filetype_from_filename_or_mimetype(data.filename, mimetype)
+    filepath = await store_file_to_disk(data.file, backend.tmpfile_dir, filetype)
```
onlyname0, line 107: Same leak on the upload side: `store_file_to_disk` writes `filepath`, `update_from_file`
consumes it, and nothing removes it afterwards. Needs a `try/finally` around the
`update_from_file` call (or cleanup inside the service once it's done reading).

`updates.py:44` has the identical problem with `store_request_stream_to_disk`.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -423,3 +476,201 @@ def to_domain(self):
...
+
+class AttributeTypeIn(InModel[AttributeType]):
+    name: AttributeNameStr
+    data_type: DataTypeIn
+    unit: t.Annotated[str, Field(max_length=ATTRIBUTE_UNIT_MAX_LENGTH)] = ""
+    description: t.Annotated[str, Field(max_length=ATTRIBUTE_DESCRIPTION_MAX_LENGTH)] = ""
+    enum_name: (
+        t.Annotated[
+            str,
+            Field(max_length=ATTRIBUTE_ENUM_NAME_MAX_LENGTH, pattern=re.compile(r"[a-z][a-z_]*")),
```
onlyname0: **This pattern is unanchored, so it accepts almost anything.** Pydantic v2 applies `pattern`
with *search* semantics, so any string merely *containing* a lowercase run passes. Posting to
`/attribute_types`:

| `enum_name` | Result |
|---|---|
| `"abc DEF"` | **200 OK, stored** |
| `"has space"` | **200 OK, stored** |
| `"Not Snake Case"` | **500** (see the `db_types.py` comment) |
| `"UPPER"` | 422 (correct) |

So values with a space or an uppercase tail are silently persisted in violation of the
constraint this is meant to enforce. Every other name field in the codebase goes through the
properly anchored `snake_case_pattern = re.compile(r"^[a-z_][a-z0-9_.]*$")`; `enum_name` is the
only hand-rolled one.

Anchoring it here would push the line over the 99-char limit, so I'd suggest a named constant
next to `snake_case_pattern` in `database/model.py`:

```python
enum_name_pattern = re.compile(r"^[a-z][a-z_]*$")
```

and then `pattern=enum_name_pattern` both here and on the column at `model.py:273`.

packages/data-core/src/movici_data_core/database/db_types.py
onlyname0: `re.match` anchors at the start but not the end, so this validates a *prefix* rather than the
whole value — `"abc DEF"` passes. Worth using `fullmatch` so that every pattern passed to
`RegexMatchingString` is safe by default, regardless of whether the caller remembered to anchor it:

```suggestion
        if not self.pattern.fullmatch(value):
```

There's a second problem here: when this check *does* fire it raises `ValueError`, which
SQLAlchemy wraps into `StatementError`. That isn't a `MoviciDataError`, so it bypasses the
handler in `make_app` and surfaces as a 500. `enum_name="Not Snake Case"` passes pydantic
(search semantics find `"ot"`) and then fails here, giving:

```
sqlalchemy.exc.StatementError: (builtins.ValueError) value Not Snake Case did not match the required pattern
```

Anchoring the pydantic pattern fixes this particular route to a 500, but the general shape —
a bind-param `ValueError` reaching the client as an internal error — is worth handling, either
by raising a `MoviciValidationError` here or by mapping `StatementError` in the app's handlers.

packages/data-core/pyproject.toml
```diff
@@ -29,17 +29,25 @@ classifiers = [
 ]
 
 dependencies = [
+  "fastapi",
```
onlyname0: `dependencies.py` uses `Depends(get_backend, scope="function")`, and `scope` is a recent
addition to FastAPI — older releases raise
`TypeError: Depends() got an unexpected keyword argument 'scope'` at import time. The lockfile
resolves 0.136.1 so CI is green, but an unbounded `"fastapi"` means anyone installing the
published package against an older release gets an import error. Could we add a lower bound
matching what we actually test against?

packages/data-core/pyproject.toml
```diff
@@ -29,17 +29,25 @@ classifiers = [
...
   "orjson>=3.11.7",
-  "sqlalchemy",
-  "movici-simulation-core",
   "pydantic>=2.7",
+  "python-multipart>=0.0.18", # from fastapi[standard]
+  "sqlalchemy",
 ]
 [project.optional-dependencies]
-sqlite = ["aiosqlite>=0.22.1"]
+standard = ["aiosqlite>=0.22.1", "fastapi-cli[standard-no-fastapi-cloud-cli]"]
```
onlyname0, line 42: Renaming the `sqlite` extra to `standard` is a silent breaking change for anyone already
installing `movici-data-core[sqlite]` — pip warns about unknown extras rather than failing, so
they'd quietly end up without `aiosqlite`. Worth a changelog entry at minimum.

packages/data-core/src/movici_data_core/api/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+            status_code=500,
+        )
+
+    for router in routers:
+        app.include_router(router)
+    return app
+
+
+def make_default_app():
+    tmpfile_path = pathlib.Path(tempfile.mkdtemp(prefix="movici_api_tmp_"))
```
onlyname0: `tempfile.mkdtemp()` here is never cleaned up, so each process start leaves a directory behind
(and with the leak above, a populated one). Since `make_app` already has a lifespan, this could
be created and torn down there instead.

Also worth noting `movici.db` on the next line resolves relative to the process working
directory, which makes the default app's database location depend on where it was launched from.

packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -0,0 +1,66 @@
...
+    UpdateWithDataOut,
+)
+from movici_simulation_core.types import FileType
+
+from .dependencies import DepBackend, DepScenarioBackend
+
+update_router = APIRouter(prefix="/updates")
+
+
+def request_filetype(content_type: t.Annotated[str | None, Header()] = None) -> FileType:
```
onlyname0, line 26: This picks the *response* format from the request's `Content-Type`, but for a GET there's no
request body to describe — `Accept` is the header that carries this. `datasets.py:63` has a TODO
acknowledging the same gap, so possibly worth solving both together.

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@dataset_router.delete("/{dataset_id}")
+async def delete_dataset(dataset_id: UUID, backend: DepBackend) -> OperationSuccess:
+    await backend.datasets.delete(dataset_id)
+    return OperationSuccess.for_path_operation(resource="dataset", id=dataset_id, verb="deleted")
+
+
+@dataset_router.get("/{dataset_id}/data")
+async def get_dataset_data(dataset_id: UUID, backend: DepBackend):
```
onlyname0: `get_dataset_data`, `create_dataset_data` (line 71) and `delete_dataset_data` (line 87) have no
return annotation, so they show up untyped in the generated OpenAPI schema while every sibling
endpoint is typed. The latter two return `OperationSuccess` already; `get_dataset_data` could use
the same `t.cast(..., FileResponse(...))` trick you used in `updates.py:66`.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -269,14 +317,14 @@ class DatasetWithDataOut(ShortDatasetOut):
     """Full output dataset model, only relevant for `ENTITY_BASED` and `UNSTRUCTURED` datasets"""
 
     epsg_code: int | None = None
-    bounding_box: BoundingBoxField | None = None
+    bounding_box: BoundingBoxField = None
```
onlyname0, line 319: Latent trap: the default is `None`, but `BoundingBoxField`'s `PlainSerializer` calls
`bbox.as_tuple_or_none()` unconditionally, so serializing a `None` raises:

```
PydanticSerializationError: Error calling function `<lambda>`:
AttributeError: 'NoneType' object has no attribute 'as_tuple_or_none'
```

Not reachable today — every domain object defaults to `BoundingBox.empty()`, which is what saves
it — but the `= None` default here invites exactly that. Either drop the default or make the
serializer None-safe (`lambda bbox: bbox.as_tuple_or_none() if bbox else None`).

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -269,14 +317,14 @@ class DatasetWithDataOut(ShortDatasetOut):
...
     general: dict | None = None
     data: dict
 
 
-class UpdateModelIn(BaseModel):
-    name: str
-    type: str | None = None
+class UpdateModelIn(InModel[UpdateModel]):
+    name: NameStr
+    type: NameStr | None = None
```
onlyname0, line 326: Same shape of trap, and this optional `type` is the line that makes it reachable:
`UpdateModelOut.type` (line 338) is `t.Annotated[str, BeforeValidator(lambda v: v.name)]`, which
raises `AttributeError` on `None`. So an `UpdateModel` built from here with `type=None` would
blow up on serialization. Only the non-nullable `update.model_type_id` FK keeps that off the read
path today. A None-safe validator would make the guarantee explicit rather than incidental.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -116,22 +151,22 @@ class ShortDatasetOut(OutModel[domain_model.Dataset]):
...
+class DatasetListOut(OutModel[t.Sequence[Dataset]]):
     __envelope__ = "datasets"
     datasets: list[ShortDatasetOut]
 
 
 class ScenarioIn(BaseModel):
-    name: str
-    display_name: str
-    description: str = ""
+    name: NameStr
```
onlyname0, line 159: nit: `ScenarioIn` (line 159) is a plain `BaseModel` while every other input model derives from
`InModel[T]` — intentional, or an oversight from the consistency pass?

packages/data-core/src/movici_data_core/api/dependencies.py
```diff
@@ -0,0 +1,69 @@
...
+        yield backend
+
+
+async def get_workspace_backend(
+    backend: DepBackend, workspace_q: t.Annotated[str | None, Query(alias="workspace")] = None
+):
+    if backend.single_workspace_mode:
+        return backend
+
+    # TODO: Validate authorization
```
onlyname0, line 30: Not for this PR, but flagging so it's tracked rather than living only in a comment: with
`GET/PUT/DELETE /datasets/{id}` and `/scenarios/{id}` on the unscoped `DepBackend`, knowing a
UUID grants access regardless of workspace. That's the security boundary, so it'd be good to
have an issue against it before this ships anywhere public.

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: hmm that seems like a reasonable approach. At least at first. We may need something else/additional for the data engine, which hands of the response to nginx ("here is a file, please send it out as a response") to limit the time spent in the application. Instead of sending the full FileReponse, we set a specific header pointing to a file; nginx reads the header and serves out that file itself, leaving FastAPI to do other things. We'd not want the background task to run in that case. However, we also don't want to return the `FileResponse` in the path function, so we'd need a different approach anyway :D

packages/data-core/src/movici_data_core/marshalling.py
elfjes: good catch :)

packages/data-core/src/movici_data_core/database/db_types.py
elfjes: changing to fullmatch here :)

as for the other issue, the current implementation is in line with other fields in sqlalchemy that raise a generic error when provided with invalid inputs. Deviating here would only touch a very minor part of that surface at the cost of inconsistency. We should make sure that our input validation is correct at the fastapi layer. The db validation only acts as a last line of defense

packages/data-core/pyproject.toml
elfjes: sure :)

packages/data-core/pyproject.toml
elfjes: We don't have a changelog yet, and we also have not released movici-data-core yet

packages/data-core/src/movici_data_core/api/__init__.py
elfjes: hmm, you make a good point. However, I think the right place to handle this is inside SQLAlchemyServer in it's `begin` context manager. So putting it there :)

packages/data-core/src/movici_data_core/api/updates.py
elfjes: Good point. However, I realize that I want to keep dealing with the accept header from this PR. Instead, we will always return json for now (where it makes sense). Removing the `filetype` dependency from the `get_update` path function for now

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: Nice catch, adding return types for create and delete dataset data is straightforward. The return type for get_dataset_data is a bit trickier though, since it may be json or arbitrary data. I'm adding an additional response type of application/octet-stream to catch the arbitrary data

packages/data-core/src/movici_data_core/marshalling.py
elfjes: sure :+1:

packages/data-core/src/movici_data_core/marshalling.py
elfjes: type=None is only allowed for UpdateModelIn and indicates that we may infer the model type from the ScenarioModel if the input does not supply it. However, as you've noted, the data model requires every update to have an associated model type so UpdateModelOut is sure to have a type.

packages/data-core/src/movici_data_core/marshalling.py
elfjes: an oversight, thanks for the catch :)

packages/data-core/src/movici_data_core/api/dependencies.py
elfjes: Since Movici Viewer will not ship with any authorization, there is a valid use case for not having authorization implemented. Only for the data engine/platform do we require it. But sure, let's create an issue for it :)

PR #144, opened 2026-08-06
Data Core: Default protected attributes

elfjes:
Describe your changes
--------------------
Depends on #143 

* creates a set of attributes that are are created by default and cannot be deleted

Copyright license
-----------------


packages/data-core/src/movici_data_core/database/repository/general.py
```diff
@@ -241,11 +246,20 @@ async def update(self, id: UUID, obj: AttributeType):
             unit=obj.unit,
             description=obj.description,
             enum_name=obj.enum_name,
+            protected=obj.protected,
```
onlyname0: **This line lets any `PUT` clear the `protected` flag, which defeats the whole feature.**

`AttributeTypeIn` has no `protected` field, so `AttributeTypeIn.to_domain()` always produces an
`AttributeType` with `protected=False`. Writing `obj.protected` into the update payload therefore
*unsets* protection on every update that comes through the API. The rename guard on line 231
doesn't help, because it only fires when the name changes — so a `PUT` that keeps the name sails
through and strips the flag:

```
DELETE /attribute_types/{id}   -> 400  Cannot delete protected attribute type 'id'
PUT    /attribute_types/{id}   -> 200  (name unchanged, description="clobbered")
DELETE /attribute_types/{id}   -> 200  attribute_type deleted

'id' still present? -> False
```

Two ordinary calls and the `id` attribute type is gone. The same two-step also defeats the rename
guard, since the second `PUT` is no longer operating on a protected row.

Simplest fix is to not let `update` write this column at all — `protected` isn't something the
API should be able to set, so dropping the line leaves the stored value untouched:

```suggestion
```

If it does need to be writable later, it should carry `current.protected` forward rather than
whatever the request body implied.

Worth noting the two new tests both drive the repository directly, which is why this slipped
through — an API-level test doing GET → PUT → DELETE would have caught it.

packages/data-core/src/movici_data_core/database/repository/general.py
```diff
@@ -241,11 +246,20 @@ async def update(self, id: UUID, obj: AttributeType):
...
         )
         await self.session.execute(
             update(db.AttributeType).where(db.AttributeType.id == id).values(**payload)
         )
 
+    async def delete(self, id: UUID):
+        current = await self.get_by_id(id)
+        if current is None:
+            return
```
onlyname0: This early `return` short-circuits the inherited `@ensure_valid_id`, so deleting a non-existent
attribute type now reports success instead of 404. It's the only resource that behaves this way:

```
DELETE /attribute_types/{random-uuid}  -> 200  {"message": "attribute_type deleted"}
DELETE /entity_types/{random-uuid}     -> 404
DELETE /dataset_types/{random-uuid}    -> 404
DELETE /model_types/{random-uuid}      -> 404
```

Raising keeps it in line with its siblings and preserves the parent's contract:

```suggestion
            raise ResourceDoesNotExist("attribute_type", id=id)
```

packages/data-core/src/movici_data_core/database/general.py
```diff
@@ -63,6 +65,79 @@ async def initialize_database(session: AsyncSession, mode: DatabaseMode):
             **_default_flags(mode),
         )
     )
+    await create_default_attribute_types(session)
+
+
+async def create_default_attribute_types(session: AsyncSession):
+    default_attributes = [
```
onlyname0: These eight are hand-transcribed from `movici_simulation_core.attributes`, which already defines
them as `AttributeSpec` objects — and `AttributeType.from_attribute_spec()` exists precisely to
convert them. I checked all eight against the canonical specs and the data types match today
(`id`=int, `geometry.x`/`y`=float, `linestring_2d`=(2,) csr, `linestring_3d`=(3,) csr,
`polygon`/`polygon_2d`=(2,) csr, `polygon_3d`=(3,) csr), but nothing keeps them that way — a
change to a spec upstream won't reach this list.

Deriving the shape/type from the specs and keeping only the descriptions and units local would
remove that drift risk.

Also: `geometry.z` is a canonical spec (`Geometry_Z`) but isn't in this list, while `x` and `y`
are. Deliberate, or an oversight?

packages/data-core/src/movici_data_core/domain_model.py
```diff
@@ -130,6 +132,7 @@ class AttributeType:
     unit: str = ""
     description: str = ""
     enum_name: str | None = None
+    protected: bool = False
```
onlyname0, line 135: `protected` isn't surfaced in `AttributeTypeOut`, so API clients can't tell which attribute types
are protected — they find out by getting a 400 back from a delete. Exposing it read-only would
also make the round-trip hazard visible: right now a client that does GET → edit → PUT silently
drops the flag, which is the mechanism behind the `repository/general.py:249` comment.

packages/data-core/src/movici_data_core/database/repository/general.py
elfjes: good catch :)

packages/data-core/src/movici_data_core/database/repository/general.py
elfjes: thanks :)

packages/data-core/src/movici_data_core/database/general.py
elfjes: makes sense, refactoring to use the AttributeSpecs

geometry.z is not exactly required as a protected attribute since we're not using it anywhere in data core, but it can't hurt to include it for consistency

packages/data-core/src/movici_data_core/domain_model.py
elfjes: sure, can't hurt to add it to AttributeTypeOut

PR #145, opened 2026-08-06
Data Core: Cache AttributeSchema in SQLAlchemyServer until invalidated

elfjes:
Describe your changes
--------------------
we don't want to retrieve the attributes for the `AttributeSchema` every time we need it, but want to cache it until it changes. This PR does that 

Depends on #143 

Copyright license
-----------------


packages/data-core/src/movici_data_core/services/common.py
```diff
@@ -40,6 +40,30 @@ async def delete(self, id: UUID):
...
+        result = await self._repository.update(id=id, obj=obj)
+        if self.invalidate_schema is not None:
+            self.invalidate_schema()
+        return result
+
+    async def delete(self, id: UUID):
+        result = await self._repository.delete(id)
+        if self.invalidate_schema is not None:
+            self.invalidate_schema()
+        return result
```
onlyname0: `make lint` fails on this file — there are no blank lines between the end of `DatasetSchemaService`
and `async def ensure_valid_workspace` below, so the module-level function reads as though it's
part of the class:

```
$ uv run ruff format --check packages/data-core/src/movici_data_core/services/common.py
Would reformat: packages/data-core/src/movici_data_core/services/common.py
1 file would be reformatted
```

CI runs `make lint`, which runs `ruff format --check`, so this fails the build as-is.
`uv run ruff format` fixes it.

packages/data-core/src/movici_data_core/services/common.py
```diff
@@ -40,6 +40,30 @@ async def delete(self, id: UUID):
         return await self._repository.delete(id)
 
 
+class DatasetSchemaService(GenericService[T_dom]):
```
onlyname0: **Invalidation is hooked in one layer above where attribute types actually get created, so the
most common creation path never invalidates.**

`EntityDataProcessor.store()` calls `AttributeTypeRepository.ensure_attribute_type()`
(`repository/common.py:210`), which creates attribute types directly on the repository —
`AttributeTypeService` is never involved, so nothing calls `invalidate_schema`. That path runs on
every dataset and update data upload whenever `STRICT_ATTRIBUTE_TYPES` is off, which is the
default for both `SINGLE_SCENARIO` and `SINGLE_WORKSPACE` (see `_default_flags`).

Against a fresh `SINGLE_WORKSPACE` database:

```
cache warmed                  -> server.schema is not None: True
'probe.attr' in cached schema -> False
after auto-create             -> cache invalidated? False
attribute type in database?   -> True
serializer schema sees it?    -> False
```

So after an upload introduces a new attribute, every later request serializes against a schema
that doesn't know about it, and stays that way until some unrelated attribute-type change happens
to come through the service. Before this PR the schema was rebuilt per backend, so this couldn't
happen.

Hooking invalidation at the repository layer — where the writes actually happen — would close
both this and the `ensure_attribute_type` case in one place.

The conftest change in this PR is the same gap showing up in tests: `an_attribute_type` creates
through the repository and then has to call `default_backend.invalidate_schema()` by hand.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -151,8 +154,12 @@ def _build_backend(self, session: AsyncSession, options: db.Options):
         assert False, f"Unknown database mode {options.mode}"
 
     async def _with_serializer(self, backend: SQLAlchemyBackend):
-        schema = await backend.attribute_types.as_schema()
-        return dataclasses.replace(backend, serializer=self.serializer.with_schema(schema))
+        if self.schema is None:
```
onlyname0: **Invalidation happens before commit, so a concurrent request can repopulate the cache with
pre-commit state — and then nothing invalidates it again.**

`invalidate_schema` fires inside the service call, but `get_backend()` only commits when its
context manager exits. Any request that rebuilds the cache in that window reads from its own
session, can't see the uncommitted row, and caches the old schema. Reproduced with two concurrent
backends on one server:

```
writer invalidated (uncommitted)   -> cache is None? True
reader repopulated cache           -> cache is None? False
reader's schema has 'probe.raced'? -> False
writer committed
AFTER COMMIT cache is None?                   -> False
AFTER COMMIT cached schema has 'probe.raced'? -> False
next request's serializer has 'probe.raced'?  -> False
```

The cache ends up populated, permanently missing an attribute type that is committed in the
database. Invalidating *after* the commit instead — in `get_backend`'s `else` branch, or off a
SQLAlchemy `after_commit` event — makes the window harmless, since a reader that repopulates
early would simply be invalidated again.

Minor related point on the same lines: the `is None` check and the `await` that follows aren't
atomic, so two concurrent misses both rebuild. Harmless (the results are equivalent) but worth
knowing it isn't a single-flight cache.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -175,6 +182,12 @@ class SQLAlchemyBackend:
     scenario_service_cls: t.Type[ScenarioService] = ScenarioService
     update_service_cls: t.Type[UpdateService] = UpdateService
 
+    invalidate_schema_callable: t.Callable[[], None] | None = None
+
+    def invalidate_schema(self):
```
onlyname0: Small cleanup: this method already null-guards `invalidate_schema_callable`, but line 219 passes
the raw callable to the service, which then repeats the same `is not None` check three times in
`DatasetSchemaService`. Passing this bound method instead would let those three checks go away.

packages/data-core/src/movici_data_core/database/backend.py
```diff
@@ -337,5 +350,11 @@ def set_options(
             self.options.IMMUTABLE_WORKSPACE_NAMES = immutable_workspace_names
 
     async def update_schema(self):
+        """Update the AttributeSchema for the serializer. For short lived SQLAlchemyBackend
```
onlyname0: Worth noting in this docstring that `update_schema()` refreshes only the backend's own
serializer — it doesn't touch `SQLAlchemyServer.schema`. So after calling it, the long-lived
backend and the server cache can disagree, and the next backend built from the server will still
get the cached schema.

onlyname0 (commented):
Caching the schema on the server is worth doing, and routing the invalidation through a
`DatasetSchemaService` mixin is a clean way to express it. But I think the invalidation is hooked
in at the wrong layer and at the wrong moment, and both gaps are reachable:

1. Attribute types created during a data upload (`ensure_attribute_type`) never invalidate,
   because that path goes through the repository, not the service. The serializer then keeps
   using a schema that's missing attributes that exist in the database.
2. Invalidation fires *before* the transaction commits, so a concurrent request can repopulate
   the cache from a session that can't see the pending row. The writer commits, nothing
   invalidates again, and the stale schema sticks.

I reproduced both against a throwaway sqlite database; traces are inline.

Also, `make lint` currently fails — `ruff format --check` wants blank lines around the new class.

packages/data-core/src/movici_data_core/database/backend.py
elfjes: good points, I'm refactoring a little bit to make it more consistent

packages/data-core/src/movici_data_core/database/backend.py
elfjes: Ha! that was my intention, but apparently I forgot to do it that way halfway through 😂

packages/data-core/src/movici_data_core/database/backend.py
elfjes: Hmm this method was only used in a test and not necessary there. Removing it alltogether

packages/data-core/src/movici_data_core/services/common.py
elfjes: Hmm. While moving it to the Repository solves the issue, it does pollute the repository. But maybe that was already the case and more logic should be placed in the service layer. However, for now I'm moving it to the repository

packages/data-core/src/movici_data_core/services/common.py
elfjes: Thanks :)

I am kind of curious how it passed pre-commit though. But I guess that doesn't matter too much

PR #146, opened 2026-08-06
Data Core: Visualization views

elfjes:
Describe your changes
--------------------
Create api endpoints and database model for CRUDding Views. Views are just unstructured json objects. We don't have a server side schema (yet). So for now we'll just store the raw json

Depends on #143 

Copyright license
-----------------


onlyname0 (commented):
The View resource is wired up cleanly and consistently with the existing resources — repository,
service, marshalling and router all follow the established shape, and the test coverage of the
CRUD paths is good. Lint, format and the full suite are green (456 passed).

Two things I'd like changed before merge:

1. `unique=True` on `VisualizationView.name` makes view names globally unique, which contradicts
   the `UniqueConstraint("scenario_id", "name")` right above it. Two scenarios can't both have a
   view called `overview`.
2. The router registers `/views/` with a trailing slash, unlike every other collection endpoint,
   so `/views` only works via a 307 redirect.

The rest is docstring copy-paste and a couple of notes. One thing I couldn't tell from the diff:
`ViewOut` doesn't include `scenario_id`, so `GET /views/{id}` gives no indication which scenario
a view belongs to — deliberate, or worth adding?

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -568,3 +568,19 @@ class UpdateAttribute(Base):
...
+
+
+class VisualizationView(Base):
+    __tablename__ = "visualization_view"
+    __table_args__ = (UniqueConstraint("scenario_id", "name"),)
+    id: Mapped[uuid.UUID] = mapped_column(primary_key=True, default=uuid.uuid4)
+    scenario_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("scenario.id", ondelete="CASCADE"))
+    name: Mapped[str] = mapped_column(
+        RegexMatchingString(pattern=snake_case_pattern, length=DEFAULT_NAME_MAX_LENGTH),
+        unique=True,
```
onlyname0: **This makes view names globally unique, which defeats the composite constraint three lines up.**

`__table_args__` declares `UniqueConstraint("scenario_id", "name")` — per-scenario uniqueness —
and `ViewRepository.get_by_name` scopes its lookup by `scenario_id`, both of which say a name only
has to be unique within its scenario. But `unique=True` here adds a second, global constraint on
the column alone, and that one wins:

```
created scenarios: ['b89e9757', '0f28532a']
scenario 1: create view 'overview' -> OK
scenario 2: create view 'overview' -> ResourceAlreadyExists: view (name="overview")
scenario 1 views: ['overview']
scenario 2 views: []
```

Through the API that second call is a 409. Since view names are user-facing labels, collisions
across scenarios seem near-certain — every scenario is going to want an `overview`.

`Scenario.name` already models exactly this correctly: a `UniqueConstraint("workspace_id", "name")`
table arg and no `unique=True` on the column. Dropping this line gives views the same behaviour:

```suggestion
```

Worth adding a test for it too — the new uniqueness tests all create both views in the same
scenario, so this passes today.

packages/data-core/src/movici_data_core/api/views.py
```diff
@@ -0,0 +1,47 @@
...
+    ViewListOut,
+    ViewOut,
+)
+
+from .dependencies import DepBackend, DepScenarioBackend
+
+view_router = APIRouter(prefix="/views")
+
+
+@view_router.get("/")
```
onlyname0: Every other collection endpoint registers `""`, not `"/"`, so this is the only one that ends up
with a trailing slash:

```
/datasets       GET, POST
/scenarios      GET, POST
/updates        GET, POST, DELETE
/workspaces     GET, POST
/views/         GET, POST     <-- odd one out
```

`/views` still works, but only because Starlette redirects:

```
GET  /views    -> 307 http://testserver/views/
POST /views    -> 307 http://testserver/views/
GET  /datasets -> 200
```

The tests in this PR call `/views` and pass because `TestClient` follows redirects by default, so
the extra hop is invisible there but real for clients. Same change needed on the `post` at line 24:

```suggestion
@view_router.get("")
```

packages/data-core/src/movici_data_core/database/repository/views.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+    @map_errors(
+        (UniqueConstraintFailed, lambda id, obj: ResourceAlreadyExists("view", name=obj.name)),
+    )
+    @ensure_valid_id
+    async def update(self, id: UUID, obj: View):
+        """Update a :class:``View`` in the database
+
+        Valid fields to update are: ``name``, ``jsonschema``
```
onlyname0: Copy-paste from `ModelTypeRepository` — views have `name` and `config`, no `jsonschema`:

```suggestion
        Valid fields to update are: ``name``, ``config``
```

packages/data-core/src/movici_data_core/database/repository/views.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+from .common import SQLResourceRepository, ensure_valid_id, validated_payload
+
+
+@dataclasses.dataclass
+class ViewRepository(SQLResourceRepository):
+    """A Repository for managing visualization views.
+
+    :param scenario_id: A Scenario UUID to bind this viewRepository to. Most methods require the
```
onlyname0: Two more copy-paste leftovers in this class: `viewRepository` should be `ViewRepository` here, and
the `list` docstring on line 46 says "these views do not contain any data", which came from
`UpdateRepository` — for a view the config *is* the data, so that line reads as though `list()`
strips something out.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -568,3 +568,19 @@ class UpdateAttribute(Base):
 
     update: Mapped[Update] = relationship()
     attribute: Mapped[Attribute] = relationship()
+
+
+class VisualizationView(Base):
```
onlyname0, line 573: Adding a table means existing databases won't have it. `create_all` only runs via `setup_db()`,
and `ensure_sqlite_db_exists` only calls that when the database file doesn't exist yet — so a
database created before this PR gets no `visualization_view` table and will fail on any view
operation. `DEFAULT_SCHEMA_VERSION` also stays `"v1"`, so nothing detects the mismatch.

I realise there's no migration tooling in data-core yet and this is likely fine pre-release, but
this is the first PR to add a table after `v1` was stamped, so it might be the moment to decide
whether the version needs bumping.

packages/data-core/src/movici_data_core/database/model.py
```diff
@@ -218,7 +218,7 @@ class Dataset(Base):
     )
     dataset_type: Mapped[DatasetType] = relationship(DatasetType)
 
-    general: Mapped[dict | None] = mapped_column(JSON)
+    general: Mapped[dict | None]
```
onlyname0: These three `mapped_column(JSON)` removals (here, `jsonschema` on 389, `simulation_info` on 410)
are unrelated to views — they lean on `Base.type_annotation_map` instead. The change looks correct
and the suite is green, but it's the kind of thing that's easier to review and revert on its own
than bundled into a feature PR.

packages/data-core/src/movici_data_core/api/views.py
elfjes: thanks :)

packages/data-core/src/movici_data_core/database/repository/views.py
elfjes: thanks for the catch :)

packages/data-core/src/movici_data_core/database/repository/views.py
elfjes: thanks for the catch

packages/data-core/src/movici_data_core/database/model.py
elfjes: yes you're right. Reverting it :)

packages/data-core/src/movici_data_core/database/model.py
elfjes: Not yet, only after we've made a release of movici-data-core

packages/data-core/src/movici_data_core/database/model.py
elfjes: You're right. the unique=True on the name field should be dropped. And I've added another test :)

PR #147, opened 2026-08-06
Data Core: endpoint for reading and updating database options

elfjes:
Describe your changes
--------------------
Implement options endpoint for reading and updating database mode and options

Depends on #143 

Copyright license
-----------------


packages/data-core/src/movici_data_core/api/options.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+
+@options_router.put("")
+async def update_scenario(
+    options: OptionsIn,
+    backend: DepBackend,
+) -> OperationSuccess:
+    if options.mode is not None:
+        await backend.set_database_mode(options.mode)
```
onlyname0, line 30: **This makes `allow_in_modes` bypassable: a client that's blocked by mode gating can just change
the mode first.**

Against a database initialized in `single_workspace`:

```
mode at start                -> single_workspace
GET /workspaces (gated)      -> 400 "workspace operation is not allowed in this mode"
PUT /options mode=multiple.. -> 200
GET /workspaces after change -> 200 {"workspaces":[{"id":"a12035b7-...","name":"__default__",...
```

So the `allow_in_modes` guards on the workspace and scenario routers are now advisory — two calls
gets you past any of them. The same applies to the flags below: one unauthenticated `PUT` turns
off every integrity constraint at once:

```
PUT /options {all strict_* : false} -> 200
resulting options -> {'mode': 'multiple_workspaces', 'strict_dataset_types': False,
                      'strict_entity_types': False, 'strict_attribute_types': False,
                      'strict_model_types': False, 'strict_scenario_datasets': False, ...}
```

`set_database_mode` also has structural side effects — the upgrade path clears
`options.default_workspace`, and downgrading to `single_scenario` will *create* a default scenario
if none exists — so this isn't purely a settings toggle.

I realise there's no auth anywhere yet, so nothing here is a regression against a boundary that
currently exists. But it does change what the API is *for*: until now a client could only act
within the configured mode, and now it can reconfigure the deployment. If mode changes are meant
to be an operator action rather than a client one, this probably wants to be admin-gated, or split
so the `STRICT_*` flags are settable but `mode` is not.

packages/data-core/src/movici_data_core/api/options.py
```diff
@@ -0,0 +1,39 @@
...
+
+from movici_data_core.marshalling import OperationSuccess, OptionsIn, OptionsOut
+
+from .dependencies import DepBackend
+
+options_router = APIRouter(prefix="/options")
+
+
+@options_router.get("")
+async def get_scenarios(backend: DepBackend) -> OptionsOut:
```
onlyname0: Copy-paste from `scenarios.py` — this leaks into the generated OpenAPI, since FastAPI derives
`operationId` from the handler name:

```
GET  /options    operationId=get_scenarios_options_get     <-- reads options
GET  /scenarios  operationId=get_scenarios_scenarios_get
```

A generated client ends up with `getScenariosOptionsGet()` for reading database options.

```suggestion
async def get_options(backend: DepBackend) -> OptionsOut:
```

packages/data-core/src/movici_data_core/api/options.py
```diff
@@ -0,0 +1,39 @@
...
+        strict_entity_types=options.STRICT_ENTITY_TYPES,
+        strict_attribute_types=options.STRICT_ATTRIBUTE_TYPES,
+        strict_model_types=options.STRICT_MODEL_TYPES,
+        strict_scenario_datasets=options.STRICT_SCENARIO_DATASETS,
+        immutable_workspace_names=options.IMMUTABLE_WORKSPACE_NAMES,
+    )
+
+
+@options_router.put("")
+async def update_scenario(
```
onlyname0: Same here — currently `operationId=update_scenario_options_put`.

```suggestion
async def update_options(
```

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -666,11 +667,31 @@ class AttributeSummaryOut(OutModel[AttributeSummary]):
     max_val: bool | int | float | None
 
 
+class OptionsIn(BaseModel):
```
onlyname0, line 669: Every field is optional and `set_options` treats `None` as "leave unchanged", so `PUT /options {}`
is a no-op rather than a reset — that's `PATCH` semantics on a `PUT` verb. Worth either switching
the route to `PATCH`, or documenting that omitted fields are preserved, since a client reasonably
expects `PUT` to replace the whole resource.

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -666,11 +667,31 @@ class AttributeSummaryOut(OutModel[AttributeSummary]):
...
+    mode: DatabaseMode | None = None
+    strict_dataset_types: bool | None = None
+    strict_entity_types: bool | None = None
+    strict_attribute_types: bool | None = None
+    strict_model_types: bool | None = None
+    strict_scenario_datasets: bool | None = None
+    immutable_workspace_names: bool | None = None
+
+
+class OptionsOut(BaseModel):
```
onlyname0, line 679: `OptionsOut` omits `default_workspace` / `default_scenario`, which the `Options` table carries and
which `set_database_mode` mutates as a side effect (the upgrade paths set them to `None`). So a
client can change the mode and has no way to observe that its default workspace reference was
cleared. Worth exposing them read-only?

packages/data-core/src/movici_data_core/marshalling.py
```diff
@@ -666,11 +667,31 @@ class AttributeSummaryOut(OutModel[AttributeSummary]):
...
+    strict_entity_types: bool
+    strict_attribute_types: bool
+    strict_model_types: bool
+    strict_scenario_datasets: bool
+    immutable_workspace_names: bool
+
+
 class OperationSuccess(BaseModel):
-    id: UUID | str
+    id: UUID | str | None = None
```
onlyname0, line 690: Making this optional means `id` is now serialized as `null` on every operation that doesn't have
one, e.g. `{"id":null,"message":"options updated","result":"ok"}`. Harmless, but if you'd rather
omit the key entirely, `model_config = ConfigDict(exclude_none=True)` on this model would do it
without affecting the existing id-carrying responses.

onlyname0 (commented):
Straightforward endpoint and the marshalling is clean — making `OperationSuccess.id` optional so
non-resource operations can reuse it is a nice touch, and I checked that every existing
`for_path_operation` caller passes keywords, so reordering the signature is safe. Lint and format
are clean, 445 tests pass.

My main question is about scope rather than correctness: this is the first endpoint that
reconfigures the *database itself* rather than manipulating data in it, and it's ungated. In
particular `PUT /options {"mode": ...}` unlocks endpoints that `allow_in_modes` is gating, which
makes that gating decorative. Given authorization is still the open TODO in `dependencies.py`,
I'd like to agree on the intended boundary before this lands. Detail inline.

Everything else is small: two copy-pasted handler names that leak into the OpenAPI spec, and a
couple of shape questions.

One more thing I noticed from your own test: downgrading the mode leaves the `STRICT_*` flags
as they were, so a database downgraded to `single_scenario` keeps `strict=True` while one
*initialized* in `single_scenario` gets `strict=False` from `_default_flags`. Same mode, different
behaviour depending on history — deliberate?

packages/data-core/src/movici_data_core/api/options.py
elfjes: thanks :)

packages/data-core/src/movici_data_core/api/options.py
elfjes: thanks

packages/data-core/src/movici_data_core/api/options.py
elfjes: `allow_in_modes` is not meant for authorization, but more for maintaining consitency when in a particular mode. I'm renaming it to `available_in_modes` to prevent confusion

After switching the mode to multiple workspaces, it is intentional that the workspaces endpoints become available. The user has upgraded the database, which unlocks the multiple workspaces feature

obviously in a multi user environment, we need to make sure the user has proper authorization to do this (and many other things). Authorzation is not in the current scope.

Currently I'm targeting usage within movici-viewer, for which a user has full control over the database and its data

packages/data-core/src/movici_data_core/marshalling.py
elfjes: fair point. Switching to PATCH

packages/data-core/src/movici_data_core/marshalling.py
elfjes: No. The default workspace and default scenario are (mostly) internal only and should be restricted from client view. Only once a client has upgraded the database should the - former - default workspace be visible.

The default scenario id already leaks a bit (given that we support the "list scenarios" and "get scenario by uuid" endpoints) but that should not be considered a feature.

packages/data-core/src/movici_data_core/marshalling.py
elfjes: sure 👍

elfjes:
regarding the strict flags. Yes, this is a deliberate choice. The defaults are chosen using the following reasoning:

* for single scenario and single workspace database, these are most likely single user environments. There should be as little friction as possible for creating resources
* multi workspace databases are most likely multi user. Multiple users should agree on certain things (such as the attribute type schema and which models are available) and not just be able to add whatever data they wish, which is why the strict flags are set by default

The above distinction (single scenario and single workspace modes are most likely single user), means that upgrading a database is also done in what was originally a single user database. It would be confusing for a user that what used to be possible (eg adding datasets with new attributes) is now no longer possible in this new mode. Setting a flag to strict should in this case be a conscious action.

PR #148, opened 2026-08-06
Data core: Dataset filtering and scenario state endpoints

elfjes:
Describe your changes
--------------------
- introduces option to  filter dataset by entity-group/attribute, so to only request a partial dataset
- add endpoint for requesting a scenario state, which is a (partial) dataset state at a certain timestamp in a simulation, by taking the initial dataset and applying the updates until that timestamp

Depends on #143 

Copyright license
-----------------


packages/data-core/tests/test_dataset_state_aggregator.py
```diff
@@ -0,0 +1,383 @@
+import typing as t
+
+import numpy as np
+import pytest
+
+from movici_data_core.database import model as db
+from movici_data_core.database.repository.state_aggregator import (
```
onlyname0: **This import breaks collection of the entire data-core test suite.** The aggregator lives at
`movici_data_core.state_aggregator`; this path is where it was before the "refactor aggregator to
prepare for dataset patching" commit moved it.

```
ModuleNotFoundError: No module named 'movici_data_core.database.repository.state_aggregator'
ERROR packages/data-core/tests/test_dataset_state_aggregator.py
!!!!!! Interrupted: 1 error during collection !!!!!!
1 error in 1.65s
```

pytest exits non-zero on a collection error, so CI is red and *no* data-core test runs. It also means
the 383 lines of new aggregator tests in this file have never executed — which I think explains the
three bugs below.

```suggestion
from movici_data_core.state_aggregator import (
```

With this fixed the rest is green: `2033 passed, 1 xfailed`.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -159,6 +201,80 @@ async def _get_bounding_box(self, scenario_id: UUID):
...
+            attribute_name = attribute.attribute_type.name
+            entity_group_name = attribute.entity_type.name
+            entity_group_data = result.setdefault(entity_group_name, {})
+            if attribute_name in entity_group_data:
+                raise ValueError(
+                    f"Duplicate attribute '{attribute_name}' found in"
+                    f" entity group '{entity_group_name}'"
+                )
+
+            data = attribute.data.to_numpy(copy=copy)
```
onlyname0: **The state endpoint 500s whenever an update modifies an attribute that already exists in the
dataset.**

`to_numpy(copy=False)` returns `np.frombuffer(self.data, ...)`, and numpy makes that array
**read-only** because it's a view over the `bytes` column. The aggregator's uniform-attribute path
writes in place (`current_data["data"][indices] = update_data["data"]`), so the write fails.

End-to-end against a real sqlite db — dataset with `topology.from_node_id`, then an update that
changes it:

```
await repo.scenarios.get_state(ScenarioStateFilter.all_attributes(dataset="a_dataset", timestamp=1))
>>> ValueError: assignment destination is read-only
```

`ValueError` isn't a `MoviciDataError`, so this surfaces as a 500.

The two new state tests pass only because their update introduces `transport.capacity`, an attribute
*not* in the initial dataset — that path allocates a fresh writable array via `get_undefined_array`.
Any update that touches an existing attribute hits this, and "taking the initial dataset and applying
the updates" is how the PR description describes the feature, so I think this is the main path rather
than an edge case.

The `copy` parameter you added to `to_numpy` in this PR is exactly the fix, it's just never passed —
`_add_attributes_to_aggregator` (line 249) calls `_combine_attributes(attributes)` and takes the
`copy=False` default. Making the initial state own its arrays fixes it:

```suggestion
            data = attribute.data.to_numpy(copy=True)
```

(Or thread `copy=True` through from `_add_attributes_to_aggregator`.) Worth a test that updates a
pre-existing attribute — that's the gap that let this through.

packages/data-core/src/movici_data_core/state_aggregator.py
```diff
@@ -0,0 +1,237 @@
...
+                        f"{', '.join(str(val) for val in ids[invalid])} not found in dataset"
+                    )
+                else:
+                    ids = self._get_filled_ids(ids)
+                    self._create_new_entities(entity_group, ids[invalid])
+                    indices = t.cast(np.ndarray, index[ids])
+            for attribute_name, attr_data in entity_group_data.items():
+                if attribute_name == "id":
+                    continue
+                indices, attr_data = strip_undefined(indices, attr_data)
```
onlyname0: **`indices` is reassigned inside the attribute loop, so one attribute's stripped indices leak into
every attribute after it.**

`strip_undefined` returns a shortened `indices` when it drops undefined entries, and that shortened
array is bound back to the loop variable rather than a local. The next attribute in the same entity
group then gets indices that no longer match its data.

With a 4-entity update where `attr_a` has two undefined holes and `attr_b` has none:

```
attr_a first:  ValueError: shape mismatch: value array of shape (4,)
               could not be broadcast to indexing result of shape (2,)
attr_b first:  b=[11, 22, 33, 44]  a=[10, 99, 30, 98]   <- correct
```

So it's an order-dependent crash — identical data succeeds or fails depending on the key order of the
incoming JSON, which is not something a client controls meaningfully. A separate name keeps each
attribute's stripping local:

```suggestion
                attr_indices, attr_data = strip_undefined(indices, attr_data)
```

(and use `attr_indices` in the `_update_attribute_data` call below).

packages/data-core/src/movici_data_core/state_aggregator.py
```diff
@@ -0,0 +1,237 @@
...
+    else:
+        if not np.any(undefs):
+            return indices, attr_data
+        return indices[~undefs], {"data": data_array[~undefs]}
+
+
+def is_undefined(arr, undefined):
+    result = arr == undefined
+    if not isinstance(undefined, str) and np.isnan(undefined):
+        return result | np.isnan(arr)
```
onlyname0: **This early return skips the reduction below, so multi-dimensional float attributes are never
reduced to one flag per entity.**

The int path falls through to `np.minimum.reduce(...)` and returns shape `(N,)`; the NaN path returns
the raw elementwise mask, shape `(N, *unit_shape)`:

```
float arr shape (3, 2) -> is_undefined shape (3, 2)
int   arr shape (3, 2) -> is_undefined shape (3,)
```

`strip_undefined` then does `indices[~undefs]` with a 2-D mask against 1-D indices:

```
2-D float, one row all-nan      -> IndexError: too many indices for array:
                                   array is 1-dimensional, but 2 were indexed
2-D int, one row all-undefined  -> OK
```

This hits real geometry — `geometry.linestring_2d`/`_3d` and the polygon attributes are all float with
a non-empty unit shape. The CSR form fails too, just differently
(`ValueError: cannot assign slice of shape (2, 2) from input of shape (1, 2)`).

Reducing on both paths fixes it — e.g. compute `result = result | np.isnan(arr)` and fall through to
the shared `np.minimum.reduce`, so the "undefined in all its dimensions" rule in the comment applies
to floats too.

packages/data-core/src/movici_data_core/api/updates.py
```diff
@@ -57,10 +60,14 @@ async def delete_updates(backend: DepScenarioBackend) -> OperationSuccess:
 
 @update_router.get("/{update_id}")
 async def get_update(
-    update_id: UUID, backend: DepBackend, filetype: DepContentType
+    update_id: UUID,
+    backend: DepBackend,
+    dataset_filter: t.Annotated[DatasetFilterIn, fastapi.Query()] | None = None,
```
onlyname0, line 71: Dropping `filetype: DepContentType` here removes the ability to fetch an update as msgpack — the
response is now hardcoded to JSON, and `UpdateService.get_update_as_file` still takes a `filetype`
that nothing passes. `DepContentType` is still used by `create_update`, so this is an asymmetry:
updates can be *uploaded* as msgpack but only *retrieved* as JSON. Intentional as part of the filter
work, or lost in the rebase?

packages/simulation-core/src/movici_simulation_core/csr.py
```diff
@@ -183,7 +183,7 @@ def update_csr_array(
         old_row = get_row(data, row_ptr, pos)
         new_row = get_row(upd_data, upd_row_ptr, upd_idx)
 
-        if changes is not None:
+        if changes is not None and compare is not None:
```
onlyname0: Making `compare` optional means `changes` is now silently ignored when a caller passes `changes=` but
no `compare=` — previously that was a `TypeError`. Since `changes` is an out-parameter, a caller that
gets this wrong sees "nothing changed" rather than an error, which is a nasty failure mode. Would
raising when `changes is not None and compare is None` be better than skipping?

packages/data-core/src/movici_data_core/state_aggregator.py
```diff
@@ -0,0 +1,237 @@
...
+
+            index, current_entity_data = self._state[entity_group]
+
+            ids = entity_group_data["id"]["data"]
+            indices = t.cast(np.ndarray, index[ids])
+
+            if np.any(invalid := (indices == -1)):
+                if not allow_new_entities:
+                    raise ValueError(
+                        f"id{'s' if len(invalid) > 1 else ''} "
```
onlyname0, line 83: `len(invalid)` is the length of the boolean mask (i.e. the total number of ids), not the number of
invalid ones, so the pluralisation is wrong whenever there's exactly one bad id among several —
"ids 5 not found in dataset". Same pattern at lines 145/147 in `_check_unique_ids`, where it also
drives the "exist/exists" suffix. `np.count_nonzero(invalid)` is what's wanted.

packages/data-core/src/movici_data_core/services/scenario.py
```diff
@@ -1,15 +1,26 @@
...
+from movici_data_core.domain_model import Scenario, ScenarioStateFilter
+from movici_data_core.exceptions import InvalidAction, ResourceDoesNotExist, UnsupportedFileType
+from movici_data_core.file_helpers import tempfile_delete_on_error
 from movici_data_core.validators import ModelConfigValidator
+from movici_simulation_core.types import ExternalSerializationStrategy, FileType
 
 
 class ScenarioService:
-    def __init__(self, repository: SQLAlchemyRepository, single_scenario_mode: bool):
+    def __init__(
```
onlyname0, line 13: `ScenarioService.__init__` gains two required positional parameters. It's referenced through
`scenario_service_cls` on the backend, which is a documented extension point, so any subclass or
alternative implementation outside this repo breaks silently at construction. Worth defaulting them
to `None`, or noting it as a breaking change.

onlyname0 (commented):
This is a big, genuinely useful piece of work — the state aggregator, the attribute filtering on
datasets/updates, and the added indices all look well thought through, and the filter-to-where-clause
approach is neat. But I don't think it's mergeable yet: the test suite doesn't collect, and once you
get past that there are three correctness bugs in the aggregator that the (unrunnable) tests would
likely have caught.

1. `tests/test_dataset_state_aggregator.py` imports the aggregator from its pre-refactor location, so
   **the whole data-core suite fails at collection** and the 383 lines of aggregator tests never run.
2. `GET /scenarios/{id}/state` **500s whenever an update modifies an attribute that already exists in
   the dataset** — i.e. the normal case for simulation results. The arrays come straight off
   `np.frombuffer`, which is read-only, and the aggregator writes into them in place.
3. `strip_undefined`'s return value is assigned back to the loop variable `indices`, so an attribute
   with undefined holes corrupts the indices used by every attribute after it in the same update.
4. `is_undefined` skips its reduction step on the NaN path, so multi-dimensional float attributes —
   including `geometry.linestring_2d` / `polygon` — crash.

All four reproduced locally; traces inline. With the collection error fixed the rest of the suite is
green (2033 passed), and lint/format are clean.

On (2): the description says the state endpoint works "by taking the initial dataset and applying
the updates until that timestamp", which is exactly the path that fails, so I don't think it's an
edge case — but do say if I've misread the intended usage.

packages/data-core/src/movici_data_core/services/scenario.py
```diff
@@ -50,3 +61,30 @@ async def get_summary(self, dataset_name_or_id: str):
         )
         assert scenario_dataset.id is not None
         return await self.repository.scenarios.get_summary(scenario_dataset.id)
+
+    async def get_state_as_file(
+        self, state_filter: ScenarioStateFilter, filetype=FileType.JSON
+    ) -> pathlib.Path | None:
+        scenario_id = self.repository.scenarios._ensure_scenario_id()
```
onlyname0: Reaching into `self.repository.scenarios._ensure_scenario_id()` from the service layer pokes at a
private method on another object. `ScenarioRepository` could expose a `scenario_id` property, or the
service could raise its own error — the current form will break quietly if the repository internals
are renamed.

packages/data-core/src/movici_data_core/api/updates.py
elfjes: (proper) support for Accept header is not part of the initial release of data core

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: I've changed the logic so that the copies are made inside `DatasetStateAggregator` and that it is no longer a concern of the consumer of the aggregator.

updated the tests to also check for an update that changes an existing attribute

packages/data-core/src/movici_data_core/services/scenario.py
elfjes: data-core is not released yet, so not worrying about breaking changes yet

packages/data-core/src/movici_data_core/services/scenario.py
elfjes: removed the call, it was only used for naming a temporary file, so not very important.

packages/data-core/src/movici_data_core/state_aggregator.py
elfjes: changing to `np.flatnonzero` which is (arguably) more readable, and also seems to perform better for larger arrays than boolean indexing

packages/data-core/src/movici_data_core/state_aggregator.py
elfjes: nice catch, thanks :)

packages/data-core/src/movici_data_core/state_aggregator.py
elfjes: thanks for the catch, added a test to verify it

packages/data-core/tests/test_dataset_state_aggregator.py
elfjes: whoops 😅

packages/simulation-core/src/movici_simulation_core/csr.py
elfjes: yes, that is better. Thanks

PR #149, opened 2026-08-06
Data core: Patch dataset endpoint

elfjes:
Describe your changes
--------------------

* implement a patch-dataset endpoint that can be used to update entities/attributes in an existing dataset without the need for uploading the full dataset
* See https://github.com/nginfra/movici-flow-lib/pull/37 for the specification of patches
  * by default, patches act like simulation updates
  * by setting `nulls_overwrite` you can deviate from the default processing of `null` values (ignore as "holes" in the update) and unset a value instead
  * new entities can be created by sending id=`null` for these entities, or a preset id (must be unique in a dataset)
  * entities can be deleted by setting the `deleted`=True (pseudo) attribute

* Depends on #143 and #144 (changes from #144 are included in this PRs diff)

Copyright license
-----------------


packages/data-core/src/movici_data_core/state_aggregator.py
```diff
@@ -90,45 +116,22 @@ def add_dataset_data(self, data: NumpyDatasetData, is_initial=False):
             for attribute_name, attr_data in entity_group_data.items():
                 if attribute_name == "id":
                     continue
-                indices, attr_data = strip_undefined(indices, attr_data)
+                if not undefined_values_overwrite:
+                    attr_data, indices = strip_undefined(attr_data, indices)
```
onlyname0: **This is why the default patch mode crashes.** `strip_undefined` returns a shortened `indices`
when it drops holes, and that gets bound back to the loop variable rather than a local — so every
attribute processed *after* one containing a null gets indices that no longer match its data.

Driving this exactly the way `patch_from_file` does (initial state, then the patch with
`undefined_values_overwrite=False`):

```
patch {"a": [9.0, null, 9.0], "b": [11,22,33]}, default mode
  -> ValueError: shape mismatch: value array of shape (3,)
     could not be broadcast to indexing result of shape (2,)

same patch with "b" before "a"
  -> OK  {'roads': {'id': [1,2,3], 'a': [9.0, 2.0, 9.0], 'b': [11.0, 22.0, 33.0]}}
```

Identical data succeeds or fails purely on JSON key order, which no client controls meaningfully.
A separate name keeps each attribute's stripping local:

```suggestion
                    attr_data, attr_indices = strip_undefined(attr_data, indices)
```

(and pass `attr_indices` to `update_attribute_data` below; the `undefined_values_overwrite=True`
branch keeps using `indices`).

packages/data-core/src/movici_data_core/state_aggregator.py
```diff
@@ -166,62 +174,108 @@ def _create_new_entities(self, entity_group: str, ids):
...
     rowptr = get_rowptr(t.cast(dict, attr_data))
-    undefined = get_undefined(attr_data["data"].dtype)
-
-    if undefined is None:
-        raise TypeError(f"unsupported dtype {data_array.dtype} ")
+    undefined_value = get_undefined_value_for_attribute(attr_data)
 
-    undefs = is_undefined(data_array, undefined)
-    num_undefined = np.sum(undefs)
+    undefs = is_undefined(data_array, undefined_value)
```
onlyname0, line 277: **Second reason the default mode breaks: `is_undefined` doesn't reduce multi-dimensional float
attributes**, so `undefs` here comes back with the wrong shape and the `data_array[~undefs]` /
`indices[~undefs]` below fail.

The int path falls through to `np.minimum.reduce` and returns one flag per entity; the NaN path
returns early with the raw elementwise mask:

```
float arr shape (3, 2) -> is_undefined shape (3, 2)
int   arr shape (3, 2) -> is_undefined shape (3,)
```

So patching any float attribute with a unit shape — which is all the geometry attributes
(`geometry.linestring_2d`/`_3d`, the polygons) — fails in default mode:

```
patch 2-D float, nulls_overwrite=False -> IndexError: too many indices for array:
                                          array is 1-dimensional, but 2 were indexed
patch 2-D float, nulls_overwrite=True  -> OK
```

Fix is in `is_undefined`: fold the NaN case into `result` and let both paths fall through to the
shared `np.minimum.reduce`, so the "undefined in all its dimensions" rule in that comment applies
to floats too.

packages/data-core/src/movici_data_core/services/dataset.py
```diff
@@ -166,8 +167,31 @@ async def update_from_file(
...
+
+        patch = DatasetPatchIn.read_from_file(path, self.serializer)
+        current_data = await self.repository.dataset_data.get_entity_data(dataset_id, copy=True)
+        aggregator = DatasetStateAggregator(allow_new_entities=True)
+        aggregator.add_dataset_data(current_data, is_initial=True)
+        aggregator.add_dataset_data(
+            patch.data, undefined_values_overwrite=patch.undefined_values_overwrite
+        )
+        await self.repository.dataset_data.delete(dataset_id)
+        await self.repository.dataset_data.create(
```
onlyname0: Patching writes the data back but never updates the dataset row, so two things go stale:

- **`bounding_box`** — the upload path recomputes it via `calculate_bounding_box_from_data` (line
  208 in this file), but patching doesn't. A patch that moves or adds geometry leaves the old box,
  which then feeds `ScenarioRepository._get_bounding_box` and `DatasetSummaryOut`.
- **`updated_at`** — it's `onupdate=func.now()`, which only fires on an `UPDATE` of `dataset`.
  Since patching only touches the attribute tables, a patched dataset still reports the timestamp
  of its last full upload.

Recomputing the box from `aggregator.state` and touching the row would cover both.

packages/data-core/tests/api/test_datasets.py
```diff
@@ -329,3 +329,38 @@ def test_get_dataset_summary(get_json, a_dataset_with_data):
             }
         ],
     }
+
+
+def test_patch_dataset(get_json, a_dataset_with_data):
+    url = f"/datasets/{a_dataset_with_data.id}/data"
+    patch = {
+        "nulls_overwrite": True,
```
onlyname0, line 363: This is the only positive patch test and it sets `nulls_overwrite: True`, which takes the branch
that skips `strip_undefined` — so the documented default (`nulls_overwrite` omitted) has no
coverage at all, which is how both crashes above got through. Worth a sibling test that omits the
flag and includes a null in a non-final attribute.

packages/data-core/src/movici_data_core/database/repository/scenario.py
```diff
@@ -252,29 +255,9 @@ def _add_attributes_to_aggregator(
         attributes: t.Iterable[db.Attribute],
         is_initial: bool,
     ):
-        dataset_data = self._combine_attributes(attributes)
+        dataset_data = combine_attributes(attributes)
```
onlyname0, line 258: Not introduced here, but this PR makes it easy to fix and the two sit next to each other now:
`combine_attributes` defaults to `copy=False`, which yields read-only `np.frombuffer` views, and
the aggregator writes into them in place. `patch_from_file` correctly passes `copy=True`; this
call doesn't, so `GET /scenarios/{id}/state` still raises `ValueError: assignment destination is
read-only` whenever an update modifies an attribute already present in the dataset. Confirmed
still failing on this branch. Passing `copy=True` here matches what the patch path already does.

packages/data-core/src/movici_data_core/api/datasets.py
```diff
@@ -92,6 +96,33 @@ async def create_dataset_data(dataset_id: UUID, data: fastapi.UploadFile, backen
...
+@dataset_router.patch("/{dataset_id}/data")
+async def patch_dataset_data(
+    dataset_id: UUID,
+    request: fastapi.Request,
+    backend: DepBackend,
+    filetype: DepContentType,
+) -> OperationSuccess:
+
+    # some checks that are also performed in the DatasetService, but we short circuit here to
+    # pfrevent additional work with handling the incoming file
```
onlyname0: Typo: "pfrevent" → "prevent". There's another in `marshalling.py:329` — "DatasetPathc" →
"DatasetPatch".

packages/data-core/src/movici_data_core/database/repository/common.py
```diff
@@ -291,7 +306,7 @@ class RawDataProcessor:
     ``UNSTRUCTURED`` datasets
     """
 
-    RAW_DATA_CHUNK_SIZE = 100_000_000  # 100 MB
+    RAW_DATA_CHUNK_SIZE = 10_000_000  # 10 MB
```
onlyname0: Dropping `RAW_DATA_CHUNK_SIZE` from 100 MB to 10 MB is unrelated to the patch feature — was that
deliberate (memory pressure?) or a stray change? Worth its own line in the description either way,
since it changes how binary datasets are chunked in storage.

packages/data-core/src/movici_data_core/api/datasets.py
elfjes: thnx!

packages/data-core/src/movici_data_core/database/repository/common.py
elfjes: you're right, reverting the change. I'm a little concerned about the 100MB chunk size. But on the other hand, decreasing it only makes sense after we've performance tested it for various sizes of datasets

packages/data-core/src/movici_data_core/database/repository/scenario.py
elfjes: this is fixed upstream in #148 by transferring the copy responsibility to the `DatasetStateAggregator`

packages/data-core/src/movici_data_core/services/dataset.py
elfjes: fixed by using DatasetRepository.update instead of DatasetDataRepository directly

packages/data-core/src/movici_data_core/state_aggregator.py
elfjes: fixed upstream in #148

packages/data-core/tests/api/test_datasets.py
elfjes: added a test 👍

